PR #4, opened 2015-12-31
Add note for debugServer setting in launch.json

gianugo:
without the debugServer switch there was no obvious way for me to debug the adapter.


felixfbecker:
Yeah I forgot to mention that, the debugServer is needed for debugging, but when you launch via "Launch extension" it _needs_ to be commented out, otherwise it will error with could not connect to port xxxx.

Issue #11, opened 2016-01-12
Remote host debugging not working


AndrewNatoli:
Hey Felix,

After getting to test with a project that isn't the included testproject I noticed that `localSourceRoot` was not being resolved properly when given an absolute path. 

`serverSourceRoot`: /var/www/
`localSourceRoot`:  /Developer/project/

See the screenshot below where the `cwd` of the adapter source directory was being combined with my `serverSourceRoot`. I imagine the actual project cwd is used when the adapter isn't being debugged?

![screen shot 2016-01-11 at 6 36 56 pm](https://cloud.githubusercontent.com/assets/4162073/12251879/4d6558dc-b8a0-11e5-85ba-a5723a3d1f1d.png)

This wasn't an issue when working with the testproject because it was relative to the adapter's source directory.

felixfbecker:
I don't get why. `path.resolve(process.cwd(), '/Developer/project')` should always return `/Developer/project`, no matter what the cwd is.... And I always though btw cwd is always the VS Code project root

felixfbecker:
@AndrewNatoli Could you please check if this issue still exists with latest master? If no, I would like to release it.

AndrewNatoli:
Hey Felix, still seems to be a no-go when using mappings with the latest master.

- Local without mapping :heavy_check_mark: 

- Local with mapping :no_entry_sign: 
*breakpoint_set sent a lowercase path and four slashes*

- Remote with mapping :no_entry_sign: 
*breakpoint_set sent with a lowercase path and four slashes*

Currently looking into what's happening....

felixfbecker:
@AndrewNatoli What settings are you using? I would like to sort this out now.
There is now a new boolean setting called `log` for launch.json that when enabled, logs all communication between VS Code and the adapter to the console even when not running in server mode. Can you please take the output you get there and post it here?

AndrewNatoli:
Linux --> OSX testing on PHP 5.3 & XDebug 2.1

Using https://github.com/felixfbecker/vscode-php-debug/commit/01e26737b934f8b137c773c3a3c4079369a3da17

## config values
```
"serverSourceRoot": "/var/www/testproject/",
"localSourceRoot": "/Developer/lserv4/testproject/"
```

## Console output 
```
-> launchResponse
Response {
  seq: 0,
  type: 'response',
  request_seq: 2,
  command: 'launch',
  success: true }

-> initializedEvent
[...]

PR #12, opened 2016-01-12
Mapping local absolute file paths

AndrewNatoli:
As mentioned in #11...

This will resolve the issue with mapping absolute file paths with the recent sourcemap additions.

If an absolute path is used, when determining the local file to open the adapter will replace the `serverSourceRoot` path with the `localSourceRoot`.

So if `serverSourceRoot` is /var/www/ and `localSourceRoot` is /Users/andrew/project/, we will take /var/www/lol.php and change it to /Users/andrew/project/lol.php.

The original logic to map relative local paths is preserved and will be used if the `localSourceRoot` begins with `.`. 

Additionally, if `serverSourceRoot` ends with a trailing slash, a trailing slash will be added to `localSourceRoot` and vice-versa to make the string.replace more convenient. 

This also ensures that `serverFileUri` is prepended with "file://" before it's sent with `set_breakpoint`.

Lastly, when parsing the uri received from XDebug (which looks like file://path/to/file.php), the original logic of removing the first character after file:// is preserved on Windows, but on non-windows systems we won't remove the first character. This will preserve /paths/like/this/.

------

Tested with....
- Remote mapping (Local = Mac, Remote = Linux)
- Local mapping
- Local (without mapping)


src/phpDebug.ts
```diff
@@ -197,14 +213,20 @@ class PhpDebugSession extends vscode.DebugSession {
 
     /** converts a server-side XDebug file URI to a local path for VS Code with respect to source root settings */
     protected convertDebuggerPathToClient(fileUri: string): string {
-        // convert the file URI to a path
-        const serverPath = decodeURI(url.parse(fileUri).pathname.substr(1));
+        // convert the file URI to a path. Don't remove starting slash on unix platforms.
+        let n:number = (/^win/.test(process.platform)) ? 1 : 0;
+        const serverPath = url.parse(fileUri).pathname.substr(n);
```
felixfbecker: This is great in general, but a check for `process.platform === 'win32'` is enough (it has nothing to do with 32 / 64 bit)

AndrewNatoli: Okay, I'll get this line revised.

felixfbecker:
I squashed my commits before I merged sourcemaps, seems like you based on my old commits. You will have to rebase interactively on upstream/master and throw my commits away. Sorry for the inconvenience.

I will revisit this tomorrow, way too tired right now (3 AM here)

felixfbecker:
In general, always start PRs of from master, not a feature branch. That feature might in the end not get merged. If it does, you can rebase before submitting ;)

felixfbecker:
And I'm a bit on the fence with this. My inital implementation used RegExps and in the end I opted for url/path resolve functions because they take care of all the inconsistencies like trailing slashes or windows drive letters in lowercase / uppercase by themselves. I would prefer if you find a solution to your issue with these tools insteads.

AndrewNatoli:
As you might be able to tell, an adventure happened during that rebase.

PR #13, opened 2016-01-12
Mapping local absolute file paths

AndrewNatoli:
Take 2.

I was able to roll back most of my changes including the `string.replace()` and end up with two minor tweaks that made the proper difference.
'
1. The platform switch on the `substring` performed on the URI returned from XDebug
2. Making sure `serverFileUri` gets "file://" prepended. 

With these in place the relative path resolving works in all of my setups.

### DIFF
--- a/src/phpDebug.ts
+++ b/src/phpDebug.ts
@@ -198,7 +198,7 @@ class PhpDebugSession extends vscode.DebugSession {
     /** converts a server-side XDebug file URI to a local path for VS Code with respect to source root settings */
     protected convertDebuggerPathToClient(fileUri: string): string {
         // convert the file URI to a path. Don't remove starting slash on unix platforms.
-        let n:number = (process.platform === 'win32') ? 1 : 0;
+        let n: number = (process.platform === 'win32') ? 1 : 0;
         const serverPath = decodeURI(url.parse(fileUri).pathname.substr(n));
         let localPath: string;
         if (this._args.serverSourceRoot && this._args.localSourceRoot) {
--- a/src/phpDebug.ts
+++ b/src/phpDebug.ts
@@ -197,8 +197,9 @@ class PhpDebugSession extends vscode.DebugSession {
 
     /** converts a server-side XDebug file URI to a local path for VS Code with respect to source root settings */
     protected convertDebuggerPathToClient(fileUri: string): string {
-        // convert the file URI to a path
-        const serverPath = decodeURI(url.parse(fileUri).pathname.substr(1));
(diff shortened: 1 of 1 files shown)


src/phpDebug.ts
```diff
@@ -227,6 +227,7 @@ class PhpDebugSession extends vscode.DebugSession {
         } else {
             serverFileUri = localFileUri;
         }
+        serverFileUri = encodeURI('file://' + serverFileUri);
         return serverFileUri;
```
felixfbecker: I think I found the actual bug. Instead of this change (and :220), in :224 `localPath` needs to be changed to `localFileUri`. In the end, I pulled url-relative in to resolve URLs, not paths :)
Please check if that is enough to fix the issue.

src/phpDebug.ts
```diff
@@ -197,8 +197,9 @@ class PhpDebugSession extends vscode.DebugSession {
 
     /** converts a server-side XDebug file URI to a local path for VS Code with respect to source root settings */
     protected convertDebuggerPathToClient(fileUri: string): string {
-        // convert the file URI to a path
-        const serverPath = decodeURI(url.parse(fileUri).pathname.substr(1));
+        // convert the file URI to a path. Don't remove starting slash on unix platforms.
+        let n:number = (process.platform === 'win32') ? 1 : 0;
```
felixfbecker: Could you add a space between `n:` and `number` please? I will add TSLint soon :)

felixfbecker: I though about this change, and I think it is not quite right. we are inspecting the OS platform of the _client_ and decide to modify the _server_ path. Isnt this dependent on the server OS rather than the client OS?

felixfbecker:
Since you were quiet for the last days I implemented the first change in 88dd1888f642.

src/phpDebug.ts
AndrewNatoli: Hey Felix,

This maps the server path returned to us to the client path. This modified path doesn't get sent back to the server so it makes sense at this point to change the path.

The reason for the different `substr` index is to keep the starting slash on Unix systems so we can map `file:///var/www` to `/var/www` (otherwise we'd have an invalid var/www). Windows won't want that starting slash so I imagine that's why you took it out initially. 

I can tell this works for linux to linux though I'm not sure how it would work for Windows to Unix or Windows to Windows.

AndrewNatoli:
Great! :smiley:

felixfbecker:
Gonna close this now as #11 is fixed :)

AndrewNatoli:
Thanks Felix!

PR #16, opened 2016-02-05
CLI Debugging instructions

elonmallin:
For those that want to debug a stand alone file. Write some simple unit test and debug it. Great to just see how the language behaves if you're new to it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,23 @@ Example:
 Both paths are normalized, so you can use slashes or backslashes no matter of the OS you're running.
 If no `localSourceRoot` is specified, the project root is assumed.
 
+CLI Debugging
+-------------
+Triggers the debugger on a single file.
+Windows (setup):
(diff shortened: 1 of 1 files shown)


felixfbecker:
This definitely is a nice addition, but I would rather have it as a launch config parameter so it can be triggered by F5. I don't really get why you have to set the IDE key - it is completely ignored by the adapter...

elonmallin:
TBH I'm not completely sure why it works, but I know it didn't work without the "SET XDEBUG_CONFIG="VSCODE". But having it in the launch config would be the way to go I agree. I just don't have any experience with creating extensions so I didn't think about that (I created a task for myself that runs a batch file with the code).

felixfbecker:
So for the record, it should work fine if you simply listen for connections and then run your script from the command line (or from a task, if you prefer).

PR #53, opened 2016-08-12
dump error messages to the debug window

bobmagicii:
so they can be fully read with line wrapping with ease.

screen shot 2016-08-12 at 5 42 08 pm


felixfbecker:
This should be enabled explicitely through launch.json (`logErrors`). This has to be added in package.json and the TS TypeScript interface. Then you can access it through `this._args`. Also please fix lint issue.

PR #55, opened 2016-08-23
support exception type error

styler2go:


felixfbecker:
Thanks for the PR. I need a test case for this, like this one: https://github.com/felixfbecker/vscode-php-debug/blob/master/src/test/adapter.ts#L199

PR #68, opened 2016-10-12
Add support for variable paging

felixfbecker:
`namedVariables` is correctly returned as being larger than the actual number of returned variables, but VS Code refuses to do any follow-up requests for paging or show any UI. I guess it is not implemented yet?


felixfbecker:
Depends on https://github.com/Microsoft/vscode/issues/13878

zobo:
Will close this as it was implemented in #805 with indexedVariables.

PR #70, opened 2016-10-12
feat: add support for setVariablesRequest

felixfbecker:
Right-click on a variable, click "set value"


DanChadwick:
:thumbsup: I would love to see this.

PR #75, opened 2016-10-20
Make max_depth and max_children configurable

batopa:
This PR add the ability to configure `max_depth` and `max_children` in launch.json file.

I left the previous values `max_depth` 5 and `max_children` 100 as default.

### DIFF
--- a/src/phpDebug.ts
+++ b/src/phpDebug.ts
@@ -60,6 +60,8 @@ interface LaunchRequestArguments extends VSCodeDebugProtocol.LaunchRequestArgume
     localSourceRoot?: string;
     /** If true, will log all communication between VS Code and the adapter to the console */
     log?: boolean;
+    /** XDebug configuration */
+    xdebugSettings?: { [key: string]: string; };
 
     // CLI options
 
@@ -130,6 +132,12 @@ class PhpDebugSession extends vscode.DebugSession {
     /** A map from unique VS Code variable IDs to XDebug eval result properties, because property children returned from eval commands are always inlined */
     private _evalResultProperties = new Map<number, xdebug.EvalResultProperty>();
 
+    /** The default XDebug settings */
+    private _defaultXdebugSettings =  {
+        maxDepth: '5',
+        maxChildren: '100'
+    };
+
     public constructor() {
         super();
         this.setDebuggerColumnsStartAt1(true);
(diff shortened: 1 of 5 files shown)


batopa:
I would like add some unit test to check that the configuration is received by xdebug but I can't figure out how to get the connection estabilished in `createServer()` function inside a test.

felixfbecker:
I don't know if this is unit-testable... we would have to send a feature_get command, but the debug adapter is always spawned in a separate process.

batopa:
Yeah, I agree on doing it more flexible to set other conf. I'm already working on it :smile: 

One question about feature names. All the feature are https://xdebug.org/docs/all_settings but for example `max_depth` is called `var_display_max_depth` in Xdebug.
So which should be used? Passing `var_display_max_depth` to `sendFeatureSetCommand()` is correct and equal to pass `max_depth`?

And about `package.json` should we just initialize the `xdebugSettings` key as `{}` in `initialConfigurations`?

felixfbecker:
Features are documented here: https://xdebug.org/docs-dbgp.php#feature-names
Not all make sense to configure of course.

src/phpDebug.ts
```diff
@@ -240,11 +242,19 @@ class PhpDebugSession extends vscode.DebugSession {
...
+                        },
+                        args.xdebugSettings || {}
+                    );
+                    for (let xdebugCommand of Object.keys(xdebugSettings)) {
+                        await connection.sendFeatureSetCommand(xdebugCommand, xdebugSettings[xdebugCommand]);
+                    }
+
                     // request breakpoints from VS Code
                     await this.sendEvent(new vscode.InitializedEvent());
                 } catch (error) {
```
batopa: If an error is triggered trying to set the feature command the debugger remains active but not respond and you need to stop it. Just an ambiguous `invalid or missing options` is visible in console.
I'd put `this.shutdown();` inside the `catch` to stop the adapter, but I'm not sure why currently it is not stopped so I left as it was.

batopa:
Fail test `should report large arrays correctly` because it expects array length of 100 (previously `max_children` I think) instead of 32 (default Xdebug?).

Suggestions?

felixfbecker:
Should either adapt the test to the XDebug default (which I don't know wether it is consistent across versions) or better set `xdebugSettings` in the tests when launching

src/phpDebug.ts
felixfbecker: Because this is the handler for a single connection, and we can have multiple connections. Just because on connection errors doesn't mean we should shutdown the whole debug adapter

batopa: Ok, I got it. Could we send an error notification maybe in addition to the output in console?

felixfbecker: There is no such thing as an error notification unfortunately, only error responses.

PR #111, opened 2016-12-06
Add DBgp proxy support

adrianofoschi:
Add support for DBgp proxy for multiuser remote debbuing.

It modifies the launchRequest adding (if proxy variables are configured) a request that register the IDE to the proxy.

See [XDebug Documentation](https://xdebug.org/docs/remote#multiple-users) for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,10 @@ In your project, go to the debugger and hit the little gear icon and choose _PHP
     - `max_data`: max amount of variable data to initially retrieve.
     - `max_depth`: maximum depth that the debugger engine may return when sending arrays, hashs or object structures to the IDE.
     - `show_hidden`: This feature can get set by the IDE if it wants to have more detailed internal information on properties (eg. private members of classes, etc.) Zero means that hidden members are not shown to the IDE.
-
+ - `proxyHost`: It is usefull for XDebug server that uses DBgp proxy to handle multiple users. Please refer to the [XDebug documentation](https://xdebug.org/docs/remote#multiple-users).
+ - `proxyPort`: See proxyHost.
+ - `ideKey`: See proxyHost.
+ 
 Options specific to CLI debugging:
  - `program`: Path to the script that should be launched
  - `args`: Arguments passed to the script
--- a/package.json
+++ b/package.json
@@ -152,6 +152,18 @@
                 "type": "boolean",
                 "description": "If true, will log all communication between VS Code and the adapter"
               },
+              "proxyHost": {
+                "type": "string",
+                "description": "DBGp Proxy host"
+              },
+              "proxyPort": {
+                "type": "number",
+                "description": "DBGp Proxy port"
+              },
+              "ideKey": {
+                "type": "string",
+                "description": "Unique key that identifies the IDE/user"
+              },
               "xdebugSettings": {
                 "type": "object",
                 "properties": {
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -50,7 +50,10 @@ In your project, go to the debugger and hit the little gear icon and choose _PHP
     - `max_data`: max amount of variable data to initially retrieve.
     - `max_depth`: maximum depth that the debugger engine may return when sending arrays, hashs or object structures to the IDE.
     - `show_hidden`: This feature can get set by the IDE if it wants to have more detailed internal information on properties (eg. private members of classes, etc.) Zero means that hidden members are not shown to the IDE.
-
+ - `proxyHost`: It is usefull for XDebug server that uses DBgp proxy to handle multiple users. Please refer to the [XDebug documentation](https://xdebug.org/docs/remote#multiple-users).
+ - `proxyPort`: See proxyHost.
+ - `ideKey`: See proxyHost.
```
felixfbecker: There should be better explanations for these settings

src/phpDebug.ts
```diff
@@ -264,8 +273,30 @@ class PhpDebugSession extends vscode.DebugSession {
...
+
+        /** sets up a client to register IDE to DBgp proxy */
+        const createProxyClient = () => new Promise((resolve, reject) => {
+            const client = this._proxyClient = net.createConnection({port: args.proxyPort | 9001, host: args.proxyHost}, () => {
+                this.sendEvent(new vscode.OutputEvent('DBgp proxy: connected to ' + args.proxyHost + ':' + args.proxyPort + '\n'));
+                client.write('proxyinit -p ' + args.proxyPort + ' -k ' + args.ideKey + ' -m 1');
+            });
+
+            client.on('data', (data:string) => {
+                client.end();
```
felixfbecker: You just end the socket on the first chunk of data, but you don't check/parse the content. This means you don't handle errors - the proxy will respond with a `proxyinit` package that can contain an error:
```xml
<proxyinit success="[0|1]"
           idekey="{ID}"
           address="{IP_ADDRESS}"
           port="{NUM}">
    <error id="app_specific_error_code">
        <message>UI Usable Message</message>
    </error>
</proxyinit>
```
The whole proxy init should not happen over a raw net socket. Instead, all the command-queue logic should be moved from `XDebugConnection` to `DBGpConnection`, and then there should be a new class `DBGpProxyConnection` that has a `sendProxyInitCommand` and `sendProxyStopCommand` method. The parsing is already implemented, that way you reduce the code here and can handle the error easily.

The `proxystop` command is currently not implemented at all.

package.json
```diff
@@ -152,6 +152,18 @@
                 "type": "boolean",
                 "description": "If true, will log all communication between VS Code and the adapter"
               },
+              "proxyHost": {
+                "type": "string",
+                "description": "DBGp Proxy host"
+              },
+              "proxyPort": {
+                "type": "number",
+                "description": "DBGp Proxy port"
```
adrianofoschi: Ok. I Will add default value.

README.md
adrianofoschi: Ok. Have you suggestions? I added the link to the xDebug documentation.

src/phpDebug.ts
adrianofoschi: Ok for the error handling.
I don't understand what do you mean with "all the command-queue logic should be moved from XDebugConnection" I worked only on PHPDebug.ts

From XDebug page I did not understand when/where proxy stop should be executed. Do you?

zobo:
Obsolete, superseded by #685

PR #121, opened 2016-12-20
Added better documentation for Linux

stanfieldr:


felixfbecker:
In CI I always use pecl to install, what do you think about this?

stanfieldr:
I added a note about it

PR #157, opened 2017-04-03
NextResponse does not take a body field, ContinueResponse does and th…

QwertyZW:
…e body takes a allThreadsContinued field

Referencing https://github.com/Microsoft/vscode-debugadapter-node/blob/master/debugProtocol.json

and the `debugProtocol.ts` generated by https://github.com/Microsoft/vscode-debugadapter-node/blob/master/debugProtocol.json

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,6 +8,7 @@ cache:
 
 environment:
   matrix:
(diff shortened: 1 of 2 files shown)


felixfbecker:
Nice catch. Could you update PHP in AppVeyor to the latest version?

QwertyZW:
Sure, will do.

QwertyZW:
As you could probably tell Xdebug 2.4.1 and below don't support php 7.1.3

PR #175, opened 2017-06-25
Path mappings

### LINKED ISSUE #54: Mapping more than one path for source code
I was wondering if it's possible to map more than one path in vscode?
I'm working with Vagrant to develop a WordPress theme, and I'm using two shared folders on the guest.
I'm sharing one folder for the root of WordPress installation called `www` and the other one which is in the same path called `main`, shared on the guest within `themes` folder of WordPress installation.

**Vagrantfile:**

```
config.vm.synced_folder "./www/", "/var/www", :mount_options => ["dmode=777", "fmode=666"]
config.vm.synced_folder "./main", "/var/www/wp-content/themes/main", :mount_options => ["dmode=777", "fmode=666"]
```

The problem is with the current launch.json file I'm not able to define two paths for my source codes, and when I'm debugging, VSCode doesn't hit the breakpoints anywhere in the `main` folder!

**Take a look at the hierarchy below:**

Folder Hierarchy

It's worth mentioning I'm currently achieving this using Sublime Text, but I'd like to work in VSCode instead.

**Sublime Text project settings:**

```
{
    "settings": {
        "xdebug": {
            "path_mapping": {
                "/var/www/wp-content/themes/main" : "/Users/[user]/Documents/[project]/main",
                "/var/www" : "/Users/[user]/Documents/[project]/www",
             },
             "super_globals": true,
             "close_on_stop": true
        }
    }
}
```

**Extra info:**

PHP version: 5.6.14-1
XDebug version: 2.3.3

Your launch.json: 

```
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Listen for XDebug",
            "type": "php",
            "request": "launch",
            "port": 9000,
            "log": true,
            "serverSourceRoot": "/var/www",
            "localSourceRoot": "/Users/[user]/Documents/[project]/www"
        },
        {
            "type": "php",
            "request": "launch",
            "program": "${file}",
            "cwd": "${fileDirname}",
            "port": 9000
        }
    ]
}
```

Thanks in advance.

Amin

johnrom:
Added Path Mapping Support via `pathMappings` option. Added Tests. Reorganized test project to support path mappings.

fixes #54


src/paths.ts
```diff
@@ -5,34 +5,87 @@ import * as url from 'url';
...
     let serverPath = decodeURI(fileUri.pathname!);
+
     // strip the trailing slash from Windows paths (indicated by a drive letter with a colon)
     const serverIsWindows = /^\/[a-zA-Z]:\//.test(serverPath);
+
     if (serverIsWindows) {
         serverPath = serverPath.substr(1);
     }
+
+    if ( pathMapping !== {} && typeof pathMapping !== 'undefined' ) {
```
aminpaks: Object references are never the same thus `{} !== {}` no?

johnrom: hm, you're correct.

aminpaks: Just check for own properties by `Object.keys` maybe?

src/paths.ts
```diff
@@ -6,34 +6,95 @@ import * as path from 'path';
 import {decode} from 'urlencode';
 
 /** converts a server-side XDebug file URI to a local path for VS Code with respect to source root settings */
-export function convertDebuggerPathToClient(fileUri: string|url.Url, localSourceRoot?: string, serverSourceRoot?: string): string {
+export function convertDebuggerPathToClient(fileUri: string|url.Url, localSourceRoot?: string, serverSourceRoot?: string, pathMapping?: { [index: string]: string; }): string {
```
felixfbecker: isn't `localSourceRoot` and `serverSourceRoot` obsolete with `pathMappings`?

src/paths.ts
johnrom: removed in d0052265ab203e67b8a38c7acd4731b6393471af

src/paths.ts
```diff
@@ -6,34 +6,95 @@ import * as path from 'path';
...
     if (typeof fileUri === 'string') {
         fileUri = url.parse(fileUri);
     }
+
     // convert the file URI to a path
     let serverPath = decode(fileUri.pathname!);
+
     // strip the trailing slash from Windows paths (indicated by a drive letter with a colon)
     const serverIsWindows = /^\/[a-zA-Z]:\//.test(serverPath);
+
```
felixfbecker: please don't try to modify the existing code style.

src/paths.ts
```diff
@@ -6,34 +6,95 @@ import * as path from 'path';
...
     if (serverIsWindows) {
         serverPath = serverPath.substr(1);
     }
+
+    if ( typeof pathMapping !== 'undefined' ) {
```
felixfbecker: please use the existing code style (no spaces inside parens)

src/paths.ts
```diff
@@ -6,34 +6,95 @@ import * as path from 'path';
...
+        let mappedServerSource: string;
+
+        for (mappedServerSource of Object.keys(pathMapping) ) {
+            let mappedLocalSource: string = pathMapping[mappedServerSource];
+
+            // normalize slashes for windows-to-unix
+            // I thought normalize did this automagically..
+            if ( process.platform !== 'win32' ) {
+                mappedServerSource = mappedServerSource.replace(/\\/g, '/');
```
felixfbecker: what if the server is Windows?

src/test/paths.ts
```diff
@@ -3,6 +3,64 @@ import {isSameUri, convertClientPathToDebugger, convertDebuggerPathToClient} fro
 import * as assert from 'assert';
 
 describe('paths', () => {
+    let undef: string;
+
+    const pathMapping = {
```
felixfbecker: I prefer to define the test cases through mocha's `it` DSL over having a config object, it's a lot easier to debug if the values are always inside the test

src/test/paths.ts
```diff
@@ -56,6 +114,43 @@ describe('paths', () => {
...
         });
+
+        describe('with path mapping', () => {
+            let sources = pathMapping.local;
+            let results = pathMapping.remote;
+
+            // unix to unix
+            it('should convert a unix path to a unix URI', () => {
+                // site
+                assert.equal( convertClientPathToDebugger(sources.unix.site, undef, undef, pathMapping.unixToUnix), results.unix.site );
```
felixfbecker: why use `undef` and not `undefined`?

src/test/adapter.ts
```diff
@@ -9,7 +9,7 @@ const assert = chai.assert;
 
 describe('PHP Debug Adapter', () => {
 
-    const TEST_PROJECT = path.normalize(__dirname + '/../../testproject');
+    const TEST_PROJECT = path.normalize(__dirname + '/../../testproject/main');
```
felixfbecker: why is this needed?

src/paths.ts
johnrom: Users that update will use "localSourceRoot" and "serverSourceRoot". It was easier to be backwards compatible with an optional extra argument, but I can certainly reformat it to inject localSourceRoot and serverSourceRoot into the config object if they exist.

(this is my first open source node contribution so I apologize for any code style differences)

src/paths.ts
aminpaks: No linting!?

src/test/adapter.ts
johnrom: Some tests use /variables.php which has been moved to /testproject/main/variables.php to test multiple source roots.

src/paths.ts
felixfbecker: BC is great, but this function should only take one source of truth for mappings

src/paths.ts
felixfbecker: There is linting, but this isn't caught.

aminpaks: Maybe turn this on => [Whitespace](https://palantir.github.io/tslint/rules/whitespace/)

felixfbecker: it is on

src/paths.ts
johnrom: This handles the case where local is Linux and server is Windows. This should be the only time backslashes need be converted because local windows will normalize the backslashes in path.relative just fine. Linux just sees the backslashes as part of the filename so path.relative returns the wrong value.

src/test/paths.ts
johnrom: ... Dunno.

src/test/adapter.ts
felixfbecker: but I don't see any new tests that rely on this?

src/paths.ts
```diff
@@ -6,34 +6,95 @@ import * as path from 'path';
...
+        for (mappedServerSource of Object.keys(pathMapping) ) {
+            let mappedLocalSource: string = pathMapping[mappedServerSource];
+
+            let localRelative: string = path.relative(mappedLocalSource, localPath);
+
+            if ( localRelative && localRelative.length > 0 ) {
+                let relative: number = localRelative.indexOf('..');
+
+                // if does not start with ..
+                if ( relative !== 0 ) {
```
felixfbecker: this is not used anywhere else, so no reason to save it in a variable. You can also use `.includes()` if I'm not mistaken.

src/paths.ts
```diff
@@ -6,34 +6,95 @@ import * as path from 'path';
...
+export function convertClientPathToDebugger(localPath: string, localSourceRoot?: string, serverSourceRoot?: string, pathMapping?: { [index: string]: string; }): string {
+
     let localFileUri = fileUrl(localPath, {resolve: false});
     let serverFileUri: string;
-    if (serverSourceRoot && localSourceRoot) {
+
+    if ( typeof pathMapping !== 'undefined' ) {
+        let mappedServerSource: string;
+
+        for (mappedServerSource of Object.keys(pathMapping) ) {
```
felixfbecker: you should declare the `let` here so it's scoped to the loop body

src/test/adapter.ts
johnrom: I'm assuming that you're requesting a test that involves `/testproject/secondary`? Please give me a little direction on this.

felixfbecker: I'm fine without a test like that given that `tests/paths.ts` unit-tests the path mapping, but I don't understand why all these files have to be moved then if they are essentially not used at all.

johnrom: when testing development of this feature, I had to test that multiple paths were working, so I created multiple paths in `testproject`. I thought that in the future other developers would also need to test multiple paths, but if you'd like I can move the files back.

felixfbecker: The testproject should only contain fixtures used by (automatic) tests

johnrom:
I believe I've made all requested changes. Let me know if you need anything else.

src/paths.ts
```diff
@@ -6,7 +6,9 @@ import * as path from 'path';
 import {decode} from 'urlencode';
 
 /** converts a server-side XDebug file URI to a local path for VS Code with respect to source root settings */
-export function convertDebuggerPathToClient(fileUri: string|url.Url, localSourceRoot?: string, serverSourceRoot?: string): string {
+export function convertDebuggerPathToClient(fileUri: string|url.Url, pathMapping?: { [index: string]: string; }): string {
+    let localSourceRoot: string = '';
+    let serverSourceRoot: string = '';
```
felixfbecker: why initialize with an empty string?

src/paths.ts
```diff
@@ -17,8 +19,23 @@ export function convertDebuggerPathToClient(fileUri: string|url.Url, localSource
     if (serverIsWindows) {
         serverPath = serverPath.substr(1);
     }
+    if (typeof pathMapping !== 'undefined') {
```
felixfbecker: shorten to `if (pathMapping)`

src/paths.ts
```diff
@@ -17,8 +19,23 @@ export function convertDebuggerPathToClient(fileUri: string|url.Url, localSource
...
+        for (let mappedServerSource of Object.keys(pathMapping) ) {
```
felixfbecker: `const`, I would name this `mappedServerPath`

src/paths.ts
```diff
@@ -17,8 +19,23 @@ export function convertDebuggerPathToClient(fileUri: string|url.Url, localSource
...
+            let mappedLocalSource: string = pathMapping[mappedServerSource];
```
felixfbecker: don't annotate types that can be inferred

src/paths.ts
```diff
@@ -17,8 +19,23 @@ export function convertDebuggerPathToClient(fileUri: string|url.Url, localSource
...
+            if (serverRelative.length > 0 && serverRelative.indexOf('..') !== 0) {
+                serverSourceRoot = mappedServerSource;
+                localSourceRoot = mappedLocalSource;
+                break;
+            }
+        }
+    }
     let localPath: string;
-    if (serverSourceRoot && localSourceRoot) {
+    if (serverSourceRoot !== '' && localSourceRoot !== '') {
```
felixfbecker: this change is unneeded

src/paths.ts
```diff
@@ -30,10 +47,24 @@ export function convertDebuggerPathToClient(fileUri: string|url.Url, localSource
...
 
 /** converts a local path from VS Code to a server-side XDebug file URI with respect to source root settings */
-export function convertClientPathToDebugger(localPath: string, localSourceRoot?: string, serverSourceRoot?: string): string {
+export function convertClientPathToDebugger(localPath: string, pathMapping?: { [index: string]: string; }): string {
+    let localSourceRoot: string = '';
+    let serverSourceRoot: string = '';
     let localFileUri = fileUrl(localPath, {resolve: false});
     let serverFileUri: string;
-    if (serverSourceRoot && localSourceRoot) {
+    if ( typeof pathMapping !== 'undefined' ) {
```
felixfbecker: remove spaces, same comment as above

src/paths.ts
```diff
@@ -30,10 +47,24 @@ export function convertDebuggerPathToClient(fileUri: string|url.Url, localSource
 }
 
 /** converts a local path from VS Code to a server-side XDebug file URI with respect to source root settings */
-export function convertClientPathToDebugger(localPath: string, localSourceRoot?: string, serverSourceRoot?: string): string {
+export function convertClientPathToDebugger(localPath: string, pathMapping?: { [index: string]: string; }): string {
+    let localSourceRoot: string = '';
+    let serverSourceRoot: string = '';
```
felixfbecker: see above

src/paths.ts
```diff
@@ -30,10 +47,24 @@ export function convertDebuggerPathToClient(fileUri: string|url.Url, localSource
...
     let localFileUri = fileUrl(localPath, {resolve: false});
     let serverFileUri: string;
-    if (serverSourceRoot && localSourceRoot) {
+    if ( typeof pathMapping !== 'undefined' ) {
+        let mappedServerSource: string;
```
felixfbecker: declare inside the loop

src/paths.ts
```diff
@@ -30,10 +47,24 @@ export function convertDebuggerPathToClient(fileUri: string|url.Url, localSource
...
+        for (mappedServerSource of Object.keys(pathMapping) ) {
+            let mappedLocalSource: string = pathMapping[mappedServerSource];
+            let localRelative: string = path.relative(mappedLocalSource, localPath);
```
felixfbecker: remove inferrable types

src/paths.ts
```diff
@@ -30,10 +47,24 @@ export function convertDebuggerPathToClient(fileUri: string|url.Url, localSource
...
+            if (localRelative.length > 0 && localRelative.indexOf('..') !== 0) {
+                serverSourceRoot = mappedServerSource;
+                localSourceRoot = mappedLocalSource;
+                break;
+            }
+        }
+    }
+    if (serverSourceRoot !== '' && localSourceRoot !== '') {
```
felixfbecker: same as above

package.json
```diff
@@ -149,6 +149,11 @@
                 "type": "string",
                 "description": "The source root on this machine that is the equivalent to the serverSourceRoot on the server."
               },
+              "pathMappings": {
+                "type": "object",
+                "default": {},
+                "description": "A mapping of multiple localSourceRoots on this machine to remote serverSourceRoots on the server."
```
felixfbecker: this shouldn't mention the name `localSourceRoot` anymore, and the old settings should be deprecated to not have two settings doing the same thing without clear precedence

src/phpDebug.ts
```diff
@@ -180,6 +182,14 @@ class PhpDebugSession extends vscode.DebugSession {
     }
 
     protected async launchRequest(response: VSCodeDebugProtocol.LaunchResponse, args: LaunchRequestArguments) {
+        if ( args.hasOwnProperty('localSourceRoot') && typeof args.localSourceRoot !== 'undefined' && args.hasOwnProperty('serverSourceRoot') && typeof args.serverSourceRoot !== 'undefined') {
```
felixfbecker: shorten to `if (args.localSourceRoot && args.serverSourceRoot)`

src/phpDebug.ts
```diff
@@ -180,6 +182,14 @@ class PhpDebugSession extends vscode.DebugSession {
...
+            let pathMappings: {[index: string]: string} = {};
+            if ( args.hasOwnProperty('pathMappings') && typeof args.pathMappings !== 'undefined') {
```
felixfbecker: same

src/paths.ts
```diff
@@ -17,8 +19,23 @@ export function convertDebuggerPathToClient(fileUri: string|url.Url, localSource
     if (serverIsWindows) {
         serverPath = serverPath.substr(1);
     }
+    if (typeof pathMapping !== 'undefined') {
+        for (let mappedServerSource of Object.keys(pathMapping) ) {
+            let mappedLocalSource: string = pathMapping[mappedServerSource];
+            // normalize slashes for windows-to-unix
+            if (process.platform !== 'win32') {
```
felixfbecker: I don't think this check is needed, this should be a pure function

src/test/paths.ts
```diff
@@ -23,37 +23,57 @@ describe('paths', () => {
...
             });
-            it('should convert a unix path to a windows URI', () => {
-                const localSourceRoot = '/home/felix/myproject';
-                const serverSourceRoot = 'C:\\Program Files\\Apache\\2.4\\htdocs';
-                assert.equal(
-                    convertClientPathToDebugger('/home/felix/myproject/test.php', localSourceRoot, serverSourceRoot),
-                    'file:///C:/Program%20Files/Apache/2.4/htdocs/test.php'
-                );
+            // unix to windows
+            (process.platform === 'win32' ? it : it.skip)('should convert a unix path to a windows URI', () => {
```
felixfbecker: if the platform check is removed, these should always run

testproject/.vscode/launch.json
```diff
@@ -7,15 +7,21 @@
...
       "name": "Launch",
       "request": "launch",
       "type": "php",
       "program": "${file}",
       "cwd": "${workspaceRoot}",
-      "externalConsole": false
+      "externalConsole": false,
+      "pathMappings": {
+        "/var/www/html": "${workspaceRoot}"
+      }
```
felixfbecker: this setting shouldn't be the default here

package.json
felixfbecker: The description says this map is local -> server, but testproject/launch.json looks like the opposite

johnrom: I'm not sure what you mean by "the old settings should be deprecated". I'm assuming I should just add "Deprecated: " to the beginning of the description?

src/paths.ts
johnrom: mappedServerSource (or mappedServerPath) is modified by the slash normalization below

src/paths.ts
johnrom: if these aren't initialized, then I receive an error testing for `serverSourceRoot && localSourceRoot` "variable is used before it is assigned"

package.json
felixfbecker: Yep, and iirc you could also set `deprecated: true`

src/paths.ts
felixfbecker: You need to type them as `string | undefined`

johnrom:
FYI, I believe I've made all requested changes here.

src/paths.ts
aminpaks: Just a friendly suggestion: tslint now has the [space within parens](https://github.com/palantir/tslint/commit/524c176684322facdb859397f9c457cc602f3349), you may enable it.

johnrom:
FYI, I believe I've made all requested changes here. Let me know if there's anything else, or at least remove the "requested changes" attribute if they have been made properly -- if that's possible.

lbod:
@johnrom thanks for this. Works a treat with my first attempt at debugging in vscode with docker wordpress/xdebug. It'll be vscode itself but the debugger response into mapped paths is very fast compared to atom.

PR #221, opened 2018-01-07
Add support for AutoHotkey

pgross41:
[AutoHotkey](https://autohotkey.com/docs/AutoHotkey.htm) uses the same debugging protocol as PHP. However, debugging an AHK script with the extension crashes. I found it is because of the "engine" attribute on line 32 of xdebugConnection.ts. AHK does not provide that attribute (note that "engine" is technically not specified in the [dbgp protocol documentation](https://xdebug.org/docs-dbgp.php#connection-initialization))

I was going to fork this into a new "AHK Debug" extension but discovered fixing that one line makes it 100% compatible with AHK. The AHK community would appreciate if you accept this request or at least implement your own code to check for the engine attribute!

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,19 +1,18 @@
 {
   "version": "0.2.0",
   "configurations": [
-    { 
-      "name": "Debug Adapter",
+    {
+      "name": "Debug adapter",
       "type": "node",
       "request": "launch",
       "program": "${workspaceRoot}/src/phpDebug.ts",
       "cwd": "${workspaceRoot}",
-      "args": ["--server=4711"], 
+      "args": ["--server=4711"],
       "env": {
         "NODE_ENV": "development"
       },
       "sourceMaps": true,
-      "outDir": "${workspaceRoot}/out",
-      // "preLaunchTask":"test"
+      "outDir": "${workspaceRoot}/out"
     },
     {
       "name": "Launch Extension",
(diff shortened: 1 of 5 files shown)


pgross41:
Ugh never mind, stepping in/over/out works, as well as the call stack, but the variables don't populate

felixfbecker:
Yeah, this adapter is focused on XDebug. I would be totally up for refactoring the DGBP implementation into a package that could be used by this adapter and your AHK adapter (vs forking). Let me know if you want to do that

pgross41:
I'm new to Node and VS Code development. Honestly it would be easier for me to fork and tweak the existing working code then implement a fresh AHK adapter. Would it be okay if I made an AHK Debug extension with my fork? I have a feeling it will only be a few more minor changes so I want to make sure you are accredited. We can also consider another pull request if the changes don't break PHP.

felixfbecker:
Well forking is always bad for OpenSource. You wont get any improvements that are done here and I wont get any improvements that are done on yours. Most code will be the exact same.

pgross41:
Yeah that makes sense. If you could refactor out the DBGP stuff as a package I would be happy to use it in a fresh AHK adapter project. Honestly I could also use xdebugConnection.ts as long as it validated the 'engine' tag exists before trying to grab the version underneath it (per this pull request). That would be much appreciated!

PR #243, opened 2018-02-23
fix(phpDebug): Make connection init more robust.

EreMaijala:
Move 'started' event after Xdebug settings have been set to make sure that 'run' is not executed before settings are processed and make sure that invalid settings don't prevent init from completing. Add some line feeds to output messages.

The main issue I stumbled on regularly was that because the 'started' event was emitted before the xdebugSettings from launch.json had been processed, occasionally the run command would be sent to Xdebug before the feature_set commands. When the feature_set commands were eventually sent, Xdebug was already in 'stopping' state and returned an error. This error was not caught and resulted in the vscode.InitializedEvent() being missed. This, in turn lead to a situation where the requests were displayed as running in the call stack window but got stuck without completing.

### DIFF
--- a/src/phpDebug.ts
+++ b/src/phpDebug.ts
@@ -290,12 +290,8 @@ class PhpDebugSession extends vscode.DebugSession {
                                 )
                             )
                         } catch (error) {
-                            this.sendEvent(
-                                new vscode.OutputEvent(
-                                    'Failed to set Xdebug settings: ' +
-                                        (error instanceof Error ? error.message : error) +
-                                        '\n'
-                                )
+                            throw new Error(
+                                'Error applying xdebugSettings: ' + (error instanceof Error ? error.message : error)
                             )
                         }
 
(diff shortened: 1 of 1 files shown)


EreMaijala:
It's easy to test the feature_set problem also by adding an invalid setting to launch.json, like this:

```
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Listen for XDebug",
            "type": "php",
            "request": "launch",
            "port": 9000,
            "xdebugSettings": {
                "invalid": true
            },
            "log": false
        },
        {
            "type": "php",
            "request": "launch",
            "program": "${file}",
            "cwd": "${fileDirname}",
            "port": 9000
        }
    ]
}
```

felixfbecker:
> make sure that invalid settings don't prevent init from completing

Unsure about this path. Why do you think is it okay to continue if the features failed? Why shouldn't we error the request (which should result in a toast vs just a log message)?

EreMaijala:
Since the settings are normally not critical for the debugger to work, I thought logging it in debug console would be better than aborting the whole process. I was also thinking that there might be other situations where setting an otherwise valid feature is not allowed for some reason beyond my knowledge, so it might be safer, but if you believe that raising a toast and erroring the request is better, I'd be fine with that too.

EreMaijala:
@felixfbecker I think this is the best I can do. The latest commit fixes startup problems e.g. when there's already a debugger running in another window and aborts the debugger if applying xdebugSettings fails. I couldn't find a way to display a nicer toast, though.

PR #244, opened 2018-02-23
Inconsistent casing

johnrom:
**Note: Merge master into this branch in order to see the real changes! I accidentally did them both at once**

This should be a complete fix for the case-sensitivity issue, though it's not the prettiest. It seems to me that either VS Code or some other dependency in my team's environment also started using lowercase drive names when reporting the file path, even though the `pathMapping` specifies an uppercase one. Breakpoints aren't working on any of my sites anymore with the current release of this extension/VSCode, which makes it borderline unusable.

- converting URIs to lowercase was not compensated for when finding the relative path between two URIs
- after converting to lowercase, a lowercase serverFileUri was returned, resulting in failed tests I chose to fix that though if returning the lowercased serverFileUri is better, that can be undone and the test can instead be changed.

I hope this is helpful.


felixfbecker:
Could you rebase this to include only your change?

johnrom:
I'm just going to build a new PR from `master`

johnrom:
closed in favor of #245

Issue #239, opened 2018-02-14
Still working after last updates from VSCode?


lmarcelocc:
All all,

I'm asking if anyone of you are having troubles after the last Visual Studio Code updates, anyone?

Thanks.

salixrosa:
Yes! Yesterday-day, I was using Xdebug and setting breakpoints and everything was perfectly fine.

Then I took my laptop home, and VS suggested it wanted to update, and I updated it, and... cue threeish hours trying to figure out what the problem was, and failing.

Eventually, I had to let it go. But my problem seemed to be somewhere down the chain of generating the serverFileUri for my breakpoints (I'm assuming that's what I see in my xdebug log). I could get VS to find the files for the exceptions XDebug gave it, but my xdebug log showed breakpoint file uris that were pretty much just the local mapping (rather than any kind of converted server mapping), and of course it did not break at the breakpoints.

I have no idea what the issue is, though, and don't have experiencing debugging extensions. Do you happen to know of an easy way of, say, logging things from the extension or launch.json to the VS console? I'd love proof of  what ${workspaceRoot} (or the other variants variables I tried out) actually outputs.

patrik-martinko:
Debugging stopped working when I updated VS Code to the v1.20. I downgraded VS Code to the v1.19.3 and debugging started working properly again.

lmarcelocc:
@salixrosa  sorry, don't know as well, how to debug the extension.

@patrik-martinko, can you share you launch.json ? Still can't get this working after downgraded to VS Code v1.19.3 :/

salixrosa:
I didn't revert my VS Code back, but I just got it working by giving up on every variation of workspaceRoot variable and doing this:

`{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Listen for XDebug",
            "type": "php",
            "request": "launch",
            "port": 9000,
            "pathMappings": {
                "/var/www": "c:/vagrant/www"
            }
        },
    ]
}`

lmarcelocc:
I wasn't use `workspaceRoot` variable before and my config is very similar to yours.

salixrosa:
What're you getting for your xdebug log?

lmarcelocc:
This is what I get when start the debugging on VS Code:
[...]

PR #198, opened 2017-10-22
Handle inconsistent casing

felixfbecker:
Closes #197

PR #245, opened 2018-02-23
Fixed Inconsistencies between Drive Paths on Windows (C:\ vs c:\)

johnrom:
It seems like VS Code or another dependency has begun using lowercase drive names in some places, but not in `${workspaceRoot}`

This will make it match up.

### DIFF
--- a/src/paths.ts
+++ b/src/paths.ts
@@ -53,7 +53,9 @@ export function convertClientPathToDebugger(localPath: string, pathMapping?: { [
     let localSourceRoot: string | undefined
     let serverSourceRoot: string | undefined
     let localFileUri = fileUrl(localPath, { resolve: false })
+    let localFileUriLower = fileUrl(localPath.replace(/[a-zA-Z]:\\/, (match: any) => match.toLowerCase()), { resolve: false })
     let serverFileUri: string
+    let uriWasLowerCased = false;
     if (pathMapping) {
         for (const mappedServerPath of Object.keys(pathMapping)) {
             const mappedLocalSource = pathMapping[mappedServerPath]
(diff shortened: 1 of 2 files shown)


felixfbecker:
Build is failing

felixfbecker:
Referencing https://github.com/felixfbecker/vscode-php-debug/issues/239

johnrom:
Prettier seems to be printing differently on my machine than in the build environment

johnrom:
Builds fixed, ready for review

src/paths.ts
```diff
@@ -53,7 +53,11 @@ export function convertClientPathToDebugger(localPath: string, pathMapping?: { [
     let localSourceRoot: string | undefined
     let serverSourceRoot: string | undefined
     let localFileUri = fileUrl(localPath, { resolve: false })
+    let localFileUriLower = fileUrl(localPath.replace(/[a-zA-Z]:\\/, (match: any) => match.toLowerCase()), {
```
felixfbecker: why is `match` `any`? Shouldn't this be inferred as `string`?

src/paths.ts
```diff
@@ -75,12 +91,20 @@ export function convertClientPathToDebugger(localPath: string, pathMapping?: { [
             serverSourceRootUrl += '/'
         }
         // get the part of the path that is relative to the source root
-        const urlRelativeToSourceRoot = urlRelative(localSourceRootUrl, localFileUri)
+        const urlRelativeToSourceRoot = urlRelative(localSourceRootUrl, localFileUriLower)
         // resolve from the server source root
         serverFileUri = url.resolve(serverSourceRootUrl, urlRelativeToSourceRoot)
+
+        // we lowercased the Url, maybe we should undo that
```
felixfbecker: This comment confuses me, what do you mean with `maybe`?

src/paths.ts
```diff
@@ -53,7 +53,11 @@ export function convertClientPathToDebugger(localPath: string, pathMapping?: { [
     let localSourceRoot: string | undefined
     let serverSourceRoot: string | undefined
     let localFileUri = fileUrl(localPath, { resolve: false })
+    let localFileUriLower = fileUrl(localPath.replace(/[a-zA-Z]:\\/, (match: any) => match.toLowerCase()), {
+        resolve: false,
+    })
     let serverFileUri: string
+    let uriWasLowerCased = false
```
felixfbecker: Frankly I am confused why this fix requires this variable - shouldn't it be possible to just _normalise_ everything to a canonical representation?

felixfbecker: E.g. before the path mapping feature I had this PR: https://github.com/felixfbecker/vscode-php-debug/pull/198
Can't the same be done, but for all paths in the map?

Issue #284, opened 2018-07-13
File not found - Mac client with Windows IIS server


Hackmodford:
PHP version:  7.0
XDebug version:  php_xdebug-2.6.0-7.0-vc14-nts.dll

launch.json
```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Debug",
            "type": "php",
            "request": "launch",
            "port": 9000,
            "pathMappings": {
                "c:\\inetpub\\mamobile dev\\mamobile": "${workspaceFolder}"
            }
        }
    ]
}
```

Problem:
VSCode cannot find files because the path is incorrect. I believe this is because I am using a mac as a client debugger with the host being a windows IIS server.

<img width="467" alt="screen-shot-2018-07-12-at-5 47 58-pm" src="https://user-images.githubusercontent.com/326668/42713071-11d13d08-86bc-11e8-81a5-c11d47dc6eaf.png">

```
/Users/hackmodford/Code/PHPStorm Projects/ma-mobile-webservice/routes\Inventory.php (incorrect path in message)
/Users/hackmodford/Code/PHPStorm Projects/ma-mobile-webservice/routes/Inventory.php (path of file in vscode)
/Users/hackmodford/Code/PHPStorm Projects/ma-mobile-webservice/routes/Inventory.php (path in FS)
```

Looks like the paths.ts needs to be fixed to address this.

Hackmodford:
This seems to be related to #216

Hackmodford:
Adding this to line 48 of paths.ts solves the problem for me. However, I am no typescript expert, so I imagine there is a better solution?

```    
if (serverIsWindows && !isWindowsUri(localPath)) {
        //Change all the backslashes to forward slashes.
        let re = /\\/gi;
        localPath = localPath.replace(re, "/");
}
```

zobo:
Closing this as it is very old and this part of code was improved and tested. Please reopen if the case is still there.

The configuration posted by OP is correct.

PR #285, opened 2018-07-14
Fix for windows server and macOS/Unix client

Hackmodford:
This is how I solved #284 

I'm not sure this is the best solution since I am no typescript expert.

### DIFF
--- a/src/paths.ts
+++ b/src/paths.ts
@@ -45,6 +45,12 @@ export function convertDebuggerPathToClient(
     } else {
         localPath = path.normalize(serverPath)
     }
+        
+    if (serverIsWindows && !isWindowsUri(localPath)) {
+        //Change all the backslashes to forward slashes if server is windows, and client is not.
+        let re = /\\/gi;
+        localPath = localPath.replace(re, "/");
+    }
     return localPath
 }
 


earth2016:
Please, this PR seems interesting to me i would like to have it for testing it.

src/paths.ts
```diff
@@ -45,6 +45,12 @@ export function convertDebuggerPathToClient(
     } else {
         localPath = path.normalize(serverPath)
     }
+        
+    if (serverIsWindows && !isWindowsUri(localPath)) {
```
felixfbecker: `localPath` is a path, not a URI, so it's not correct to use `isWindowsUri()` on it

Hackmodford: Thanks for the input. I am not sure how to fix this and mostly submitted this pull request for the discussion. It does solve the issue on my machine, but if anybody knows a better solution, by all means please help :)

lastthyme (approved):

zobo:
Is this somehow relevant? There were a number of changes in this logic (and some pending ones) and if it's still a problem I'll write tests and see about the change. Honestly I did not use it in this context (server windows, client POSIX).

zobo:
Outdated and obsolete. Using `clientIsWindows ? path.win32 : path.posix` and `serverIsWindows ? path.win32 : path.posix`.

Issue #288, opened 2018-08-01
XDebug fails to connect because vscode-php-debug is listening on IPv6


mscharley:
PHP version:  7.1.20
XDebug version:  2.6.0
Adapter version: 1.12.3

Your launch.json:  

```
{
  // Use IntelliSense to learn about possible attributes.
  // Hover to view descriptions of existing attributes.
  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
  "version": "0.2.0",
  "configurations": [
    {
      "name": "Listen for XDebug",
      "type": "php",
      "request": "launch",
      "pathMappings": {
        "/portal": "${workspaceFolder}",
      },
      "port": 9000,
      "log": true
    },
    {
      "name": "Launch currently open script",
      "type": "php",
      "request": "launch",
      "program": "${file}",
      "cwd": "${fileDirname}",
      "port": 9000
    }
  ]
}
```

XDebug php.ini config:
```
zend_extension=/usr/local/lib/php/extensions/no-debug-non-zts-20160303/xdebug.so
xdebug.remote_enable=1
xdebug.remote_autostart=1
xdebug.remote_connect_back=1
xdebug.remote_log=/var/log/xdebug.log
xdebug.remote_host=host.docker.internal
```

XDebug logfile (from setting `xdebug.remote_log` in php.ini):  
```
Log opened at 2018-08-01 02:13:59
I: Checking remote connect back address.
I: Checking header 'HTTP_X_FORWARDED_FOR'.
I: Checking header 'REMOTE_ADDR'.
I: Remote address found, connecting to 172.19.0.1:9000.
W: Creating socket for '172.19.0.1:9000', poll success, but error: Operation now in progress (29).
E: Could not connect to client. :-(
Log closed at 2018-08-01 02:13:59
```

Adapter logfile (from setting `"log": true` in launch.json):
```
<- launchResponse
Response {
  seq: 0,
  type: 'response',
  request_seq: 2,
  command: 'launch',
  success: true }
```

Code snippet to reproduce:
```php
<?php
phpinfo();
```

Also relevant:
```
$ lsof -nP -i4TCP:9000
COMMAND     PID            USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
Code\x20H 46903 matthewscharley   28u  IPv6 0x32d841ebd93c8e67      0t0  TCP *:9000 (LISTEN)
```

[...]

PR #289, opened 2018-08-01
feat: Add support for listening to a specific network interface if needed.

mscharley:
This is my attempt at fixing #288

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ In your project, go to the debugger and hit the little gear icon and choose _PHP
 #### Supported launch.json settings:
 
 - `request`: Always `"launch"`
+- `bind`: The address to bind to when listening for XDebug (default: all IPv6 connections if available, else all IPv4 connections)
 - `port`: The port on which to listen for XDebug (default: `9000`)
 - `stopOnEntry`: Wether to break at the beginning of the script (default: `false`)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -41,6 +41,7 @@ In your project, go to the debugger and hit the little gear icon and choose _PHP
 #### Supported launch.json settings:
 
 - `request`: Always `"launch"`
+- `bind`: The address to bind to when listening for XDebug (default: all IPv6 connections if available, else all IPv4 connections)
```
felixfbecker: I find this name a bit confusing. The argument to `server.listen()` is named `hostname`. Could we name it that?

mscharley: We can, I kind of find hostname confusing too in this context - it sounds like the name of the host that XDebug is connecting from. I've seen it named `bind` in other applications so that's why I used that name here. I think `hostname` makes sense in the context of a call to `server.listen()` but in the context of a configuration for this module it's ambiguous.

package.json
```diff
@@ -174,6 +174,11 @@
                 "description": "Environment variables passed to the program.",
                 "default": {}
               },
+              "bind": {
+                "type": "string",
+                "description": "Address to bind to when listening for XDebug",
+                "default": "::"
```
mscharley: Ideally this would default to null if possible, but not familiar with VS Code plugins to know if that is or not.

PR #292, opened 2018-08-14
Prevent deadlock by only updating breakpoints on needed connections

### LINKED ISSUE #164: Multiple calls make it hang
For a project, we have multiple micro services, running in different containers each set up with it's own xdebug port so I can debug a specific micro service or multiple microservices at once.  Everything is working well, breakpoints break, code is able to step properly etc.  Unless one microservice makes a call to another one that makes a call back to the first micro service, at which point it hangs until the connection times out.

Before it times out though, in the editor it shows 2 requests on the call stack.  If I put a breakpoint in the code in "container A" before the point it makes that call to container B, it breaks just fine.  If I then try to step through it, it hangs once it gets to the line that makes the call to container B until the container B call times out, then it continues on with a timeout error creating a stop point in xdebug.

I was able to simplify the error down to a simple test not involving multiple containers or anything, basically just have a call make a curl connection to make a second call.

PHP version:  5.6.30
XDebug version:  2.5.0
Adapter version:  1.10.0

Your launch.json:  
```
{
	"version": "0.2.0",
	"configurations": [
		{
			"name": "Listen for XDebug",
			"type": "php",
			"request": "launch",
			"port": 9002,
			"localSourceRoot": "${workspaceRoot}",
			"serverSourceRoot": "/var/www/public/releases/local_source"
		}
	]
}
```
XDebug php.ini config:  
```
zend_extension=xdebug.so
[xdebug]
xdebug.remote_enable=1
xdebug.remote_autostart=1
xdebug.remote_connect_back=0
xdebug.remote_mode=req
xdebug.remote_port=9002
xdebug.remote_host=10.55.55.1
```

XDebug logfile (from setting `xdebug.remote_log` in php.ini):  
```
Log opened at 2017-05-15 17:02:39
I: Connecting to configured address/port: 10.55.55.1:9002.
I: Connected to client. :-)
-> <init xmlns="urn:debugger_protocol_v1" xmlns:xdebug="http://xdebug.org/dbgp/xdebug" fileuri="file:///var/www/public/releases/local_source/app/webroot/test2.php" language="PHP" xdebug:language_version="5.6.30-1+deb.sury.org~xenial+1" protocol_version="1.0" appid="26056"><engine version="2.5.0"><![CDATA[Xdebug]]></engine><author><![CDATA[Derick Rethans]]></author><url><![CDATA[http://xdebug.org]]></url><copyright><![CDATA[Copyright (c) 2002-2016 by Derick Rethans]]></copyright></init>

<- breakpoint_list -i 1
-> <response xmlns="urn:debugger_protocol_v1" xmlns:xdebug="http://xdebug.org/dbgp/xdebug" command="breakpoint_list" transaction_id="1"></response>

<- breakpoint_set -i 2 -t line -f file:///var/www/public/releases/local_source/app/webroot/test2.php -n 10
-> <response xmlns="urn:debugger_protocol_v1" xmlns:xdebug="http://xdebug.org/dbgp/xdebug" command="breakpoint_set" transaction_id="2" id="260560003"></response>

<- breakpoint_list -i 3
-> <response xmlns="urn:debugger_protocol_v1" xmlns:xdebug="http://xdebug.org/dbgp/xdebug" command="breakpoint_list" transaction_id="3"><breakpoint type="line" filename="file:///var/www/public/releases/local_source/app/webroot/test2.php" lineno="10" state="enabled" hit_count="0" hit_value="0" id="260560003"></breakpoint></response>

<- breakpoint_list -i 4
-> <response xmlns="urn:debugger_protocol_v1" xmlns:xdebug="http://xdebug.org/dbgp/xdebug" command="breakpoint_list" transaction_id="4"><breakpoint type="line" filename="file:///var/www/public/releases/local_source/app/webroot/test2.php" lineno="10" state="enabled" hit_count="0" hit_value="0" id="260560003"></breakpoint></response>

<- breakpoint_set -i 5 -t exception -x *
-> <response xmlns="urn:debugger_protocol_v1" xmlns:xdebug="http://xdebug.org/dbgp/xdebug" command="breakpoint_set" transaction_id="5" id="260560004"></response>

<- run -i 6
Log opened at 2017-05-15 17:02:39
I: Connecting to configured address/port: 10.55.55.1:9002.
I: Connected to client. :-)
-> <init xmlns="urn:debugger_protocol_v1" xmlns:xdebug="http://xdebug.org/dbgp/xdebug" fileuri="file:///var/www/public/releases/local_source/app/webroot/test2.php" language="PHP" xdebug:language_version="5.6.30-1+deb.sury.org~xenial+1" protocol_version="1.0" appid="26052"><engine version="2.5.0"><![CDATA[Xdebug]]></engine><author><![CDATA[Derick Rethans]]></author><url><![CDATA[http://xdebug.org]]></url><copyright><![CDATA[Copyright (c) 2002-2016 by Derick Rethans]]></copyright></init>

<- breakpoint_list -i 1
-> <response xmlns="urn:debugger_protocol_v1" xmlns:xdebug="http://xdebug.org/dbgp/xdebug" command="breakpoint_list" transaction_id="1"></response>

<- breakpoint_set -i 2 -t line -f file:///var/www/public/releases/local_source/app/webroot/test2.php -n 10
-> <response xmlns="urn:debugger_protocol_v1" xmlns:xdebug="http://xdebug.org/dbgp/xdebug" command="breakpoint_set" transaction_id="2" id="260520005"></response>

<- stop -i 3
-> <response xmlns="urn:debugger_protocol_v1" xmlns:xdebug="http://xdebug.org/dbgp/xdebug" command="stop" transaction_id="3" status="stopped" reason="ok"></response>

-> <response xmlns="urn:debugger_protocol_v1" xmlns:xdebug="http://xdebug.org/dbgp/xdebug" command="run" transaction_id="6" status="stopping" reason="ok"></response>

<- breakpoint_list -i 7
-> <response xmlns="urn:debugger_protocol_v1" xmlns:xdebug="http://xdebug.org/dbgp/xdebug" command="breakpoint_list" transaction_id="7"><breakpoint type="line" filename="file:///var/www/public/releases/local_source/app/webroot/test2.php" lineno="10" state="enabled" hit_count="0" hit_value="0" id="260560003"></breakpoint><breakpoint type="exception" state="enabled" hit_count="0" hit_value="0" id="260560004"></breakpoint></response>

<- stop -i 8
-> <response xmlns="urn:debugger_protocol_v1" xmlns:xdebug="http://xdebug.org/dbgp/xdebug" command="stop" transaction_id="8" status="stopped" reason="ok"></response>

Log closed at 2017-05-15 17:02:48
```
Adapter logfile (from setting `"log": true` in launch.json):  
```
<- launchResponse
Response {
  seq: 0,
  type: 'response',
  request_seq: 2,
  command: 'launch',
  success: true }
new connection 1
<- threadEvent
ThreadEvent {
  seq: 0,
  type: 'event',
  event: 'thread',
  body: { reason: 'started', threadId: 1 } }
<- initializedEvent
InitializedEvent { seq: 0, type: 'event', event: 'initialized' }
-> threadsRequest
{ command: 'threads', type: 'request', seq: 3 }
<- threadsResponse
Response {
  seq: 0,
  type: 'response',
  request_seq: 3,
  command: 'threads',
  success: true,
  body: { threads: [ Thread { id: 1, name: 'Request 1 (1:02:39 PM)' } ] } }
-> setBreakpointsRequest
{ command: 'setBreakpoints',
  arguments: 
   { source: 
      { path: '/Users/jonathanfoote/zumba/git/public/app/webroot/test2.php',
        name: 'test2.php' },
     lines: [ 10 ],
     breakpoints: [ { line: 10 } ],
     sourceModified: false },
  type: 'request',
  seq: 4 }
<- setBreakpointsResponse
Response {
  seq: 0,
  type: 'response',
  request_seq: 4,
  command: 'setBreakpoints',
  success: true,
  body: { breakpoints: [ { verified: true, line: 10 } ] } }
-> setFunctionBreakpointsRequest
{ command: 'setFunctionBreakpoints',
  arguments: { breakpoints: [] },
  type: 'request',
  seq: 5 }
<- setFunctionBreakpointsResponse
Response {
  seq: 0,
  type: 'response',
  request_seq: 5,
  command: 'setFunctionBreakpoints',
  success: true,
  body: { breakpoints: [] } }
-> setExceptionBreakpointsRequest
{ command: 'setExceptionBreakpoints',
  arguments: { filters: [ '*' ] },
  type: 'request',
  seq: 6 }
<- setExceptionBreakpointsResponse
Response {
  seq: 0,
  type: 'response',
  request_seq: 6,
  command: 'setExceptionBreakpoints',
  success: true }
-> configurationDoneRequest
{ command: 'configurationDone', type: 'request', seq: 7 }
-> threadsRequest
{ command: 'threads', type: 'request', seq: 8 }

<- threadsResponse
Response {
  seq: 0,
  type: 'response',
  request_seq: 8,
  command: 'threads',
  success: true,
  body: { threads: [ Thread { id: 1, name: 'Request 1 (1:02:39 PM)' } ] } }
new connection 2
<- threadEvent
ThreadEvent {
  seq: 0,
  type: 'event',
  event: 'thread',
  body: { reason: 'started', threadId: 2 } }
<- initializedEvent
InitializedEvent { seq: 0, type: 'event', event: 'initialized' }
-> threadsRequest
{ command: 'threads', type: 'request', seq: 9 }
<- threadsResponse
Response {
  seq: 0,
  type: 'response',
  request_seq: 9,
  command: 'threads',
  success: true,
  body: 
   { threads: 
      [ Thread { id: 1, name: 'Request 1 (1:02:39 PM)' },
        Thread { id: 2, name: 'Request 2 (1:02:39 PM)' } ] } }
-> setBreakpointsRequest
{ command: 'setBreakpoints',
  arguments: 
   { source: 
      { path: '/Users/jonathanfoote/zumba/git/public/app/webroot/test2.php',
        name: 'test2.php' },
     lines: [ 10 ],
     breakpoints: [ { line: 10 } ],
     sourceModified: false },
  type: 'request',
  seq: 10 }
-> disconnectRequest
{ command: 'disconnect',
  arguments: { restart: false },
  type: 'request',
  seq: 11 }
<- configurationDoneResponse
Response {
  seq: 0,
  type: 'response',
  request_seq: 7,
  command: 'configurationDone',
  success: true }
<- outputEvent
OutputEvent {
  seq: 0,
  type: 'event',
  event: 'output',
  body: { category: 'console', output: 'connection 1 closed\n' } }
connection 1 closed
<- threadEvent
ThreadEvent {
  seq: 0,
  type: 'event',
  event: 'thread',
  body: { reason: 'exited', threadId: 1 } }
-> setFunctionBreakpointsRequest
{ command: 'setFunctionBreakpoints',
  arguments: { breakpoints: [] },
  type: 'request',
  seq: 12 }
```

Code snippet to reproduce:  
```php
<?php

if (empty($_GET['connect'])) {
	$ch = curl_init('http://localhost/' . $_SERVER['REQUEST_URI'] . '?connect=1');
	curl_exec($ch);
	curl_close($ch);
	die ('end of initial call');
}

phpinfo();

```

jonyo:
This fixes #164 

In this PR:
* When making calls related to updating the breakpoints, if there is a connection set in `_waitingConnections`, it only uses that connection instead of all.  When that is the case the breakpoints are likely only getting updated as part of initializing the new connection.  This fixes the main problem mentioned in #164 that was caused because it was sending the breakpoint calls to the first connection, when that connection was locked up waiting on response from second connection.  Doing it this way eliminates the deadlock, at least in this scenario.
* When sending a command, record the command in `_pendingConnections` **before** it makes the call, not after.  This ensures it correctly only attempts to send one command at a time and waits for the response before sending another.
* When sending a command that tells xdebug to continue executing PHP (like run, step_into, etc), it now sends the `ContinuedEvent`.  With the change, now once you hit continue or step into etc. it immediately behaves like the execution has continued.  Before the change if there was something like a sleep() command, when you hit continue, it would "seem" like nothing happened as it appears to still be paused on a breakpoint when that is not the case.

### DIFF
--- a/src/phpDebug.ts
+++ b/src/phpDebug.ts
@@ -274,11 +274,15 @@ class PhpDebugSession extends vscode.DebugSession {
                                 this._waitingConnections.delete(connection)
                             }
                         }
+                        const continueConnection = () => {
+                            this.sendEvent(new vscode.ContinuedEvent(connection.id));
+                        }
                         connection.on('warning', (warning: string) => {
                             this.sendEvent(new vscode.OutputEvent(warning + '\n'))
                         })
                         connection.on('error', disposeConnection)
                         connection.on('close', disposeConnection)
+                        connection.on('continue-event', continueConnection);
                         await connection.waitForInitPacket()
 
                         // override features from launch.json
@@ -437,7 +441,9 @@ class PhpDebugSession extends vscode.DebugSession {
     ) {
         try {
             const fileUri = convertClientPathToDebugger(args.source.path!, this._args.pathMappings)
-            const connections = Array.from(this._connections.values())
+            // If there is a waiting connection, only use use that as it is likely being called to init
+            // the breakpoints on the new connection
+            const connections = Array.from(this._waitingConnections.size ? this._waitingConnections.values() : this._connections.values())
             let xdebugBreakpoints: Array<xdebug.ConditionalBreakpoint | xdebug.LineBreakpoint>
             response.body = { breakpoints: [] }
             // this is returned to VS Code
@@ -510,7 +516,9 @@ class PhpDebugSession extends vscode.DebugSession {
         args: VSCodeDebugProtocol.SetExceptionBreakpointsArguments
     ) {
         try {
-            const connections = Array.from(this._connections.values())
+            // If there is a waiting connection, only use use that as it is likely being called to init
+            // the breakpoints on the new connection
+            const connections = Array.from(this._waitingConnections.size ? this._waitingConnections.values() : this._connections.values())
             await Promise.all(
                 connections.map(async connection => {
                     // get all breakpoints
@@ -541,7 +549,9 @@ class PhpDebugSession extends vscode.DebugSession {
         args: VSCodeDebugProtocol.SetFunctionBreakpointsArguments
     ) {
         try {
-            const connections = Array.from(this._connections.values())
+            // If there is a waiting connection, only use use that as it is likely being called to init
+            // the breakpoints on the new connection
+            const connections = Array.from(this._waitingConnections.size ? this._waitingConnections.values() : this._connections.values())
             // this is returned to VS Code
             let vscodeBreakpoints: VSCodeDebugProtocol.Breakpoint[]
             if (connections.length === 0) {
--- a/src/xdebugConnection.ts
+++ b/src/xdebugConnection.ts
@@ -649,8 +649,8 @@ export class Connection extends DbgpConnection {
         }
         commandString += '\0'
         const data = iconv.encode(commandString, ENCODING)
-        await this.write(data)
         this._pendingCommands.set(transactionId, command)
+        await this.write(data)
     }
 
     public close() {
@@ -752,21 +752,25 @@ export class Connection extends DbgpConnection {
 
     /** sends a run command */
     public async sendRunCommand(): Promise<StatusResponse> {
+        this.emit('continue-event');
         return new StatusResponse(await this._enqueueCommand('run'), this)
     }
 
     /** sends a step_into command */
     public async sendStepIntoCommand(): Promise<StatusResponse> {
+        this.emit('continue-event');
         return new StatusResponse(await this._enqueueCommand('step_into'), this)
     }
 
     /** sends a step_over command */
     public async sendStepOverCommand(): Promise<StatusResponse> {
+        this.emit('continue-event');
         return new StatusResponse(await this._enqueueCommand('step_over'), this)
     }
 
     /** sends a step_out command */
     public async sendStepOutCommand(): Promise<StatusResponse> {
+        this.emit('continue-event');
         return new StatusResponse(await this._enqueueCommand('step_out'), this)
     }
 


src/xdebugConnection.ts
```diff
@@ -649,8 +649,8 @@ export class Connection extends DbgpConnection {
         }
         commandString += '\0'
         const data = iconv.encode(commandString, ENCODING)
-        await this.write(data)
         this._pendingCommands.set(transactionId, command)
+        await this.write(data)
```
felixfbecker: 👍

src/xdebugConnection.ts
```diff
@@ -752,21 +752,25 @@ export class Connection extends DbgpConnection {
 
     /** sends a run command */
     public async sendRunCommand(): Promise<StatusResponse> {
+        this.emit('continue-event');
```
felixfbecker: "ContinueEvent" is a VS Code concept, but this class should have no knowledge of VS Code, only of XDebug. I would prefer if callers of these `send*Command()` functions in `PhpDebugSession` sent the `ContinuedEvent` to VS Code, without this layer of event indirection.

jonyo:
@felixfbecker 
```bash
$ npm -v
4.2.0
$ node -v
v7.9.0
$ php -v
PHP 7.1.16 (cli) (built: Mar 31 2018 02:59:59) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2018 Zend Technologies
    with Xdebug v2.5.5, Copyright (c) 2002-2017, by Derick Rethans
```

src/phpDebug.ts
```diff
@@ -437,7 +441,9 @@ class PhpDebugSession extends vscode.DebugSession {
     ) {
         try {
             const fileUri = convertClientPathToDebugger(args.source.path!, this._args.pathMappings)
-            const connections = Array.from(this._connections.values())
+            // If there is a waiting connection, only use use that as it is likely being called to init
+            // the breakpoints on the new connection
```
felixfbecker: What concerns me about this fix is the word "likely". "likely" to me means "race condition". Can this be fixed without having to rely on the likelihood of something?

I think fundamentally the problem here is that for a single _VS Code <-> debug adapter_ connection we need to persist the breakpoints to _multiple_ connected PHP processes. Some may block because they are mid-execution, some may succeed, some may fail, but we only have a single request from VS Code to handle and respond to (with verified or error).

@weinand any advice how to handle this scenario best?

felixfbecker:
Try updating npm to latest (6.3.0). If it still doesn't work, post me the error

jonyo:
@felixfbecker What version of node should I be using for this?  After updating npm it warns about using npm with this version of node, will it work with latest version of node?

```
$ npm test
npm WARN npm npm does not support Node.js v7.9.0
npm WARN npm You should probably upgrade to a newer version of node as we
npm WARN npm can't make any promises that npm will work with this version.
npm WARN npm Supported releases of Node.js are the latest release of 6, 8, 9, 10, 11.
npm WARN npm You can find the latest version at https://nodejs.org/
```

felixfbecker:
It used to be 7.9, but VS Code just updated to 8.9.3 https://github.com/felixfbecker/vscode-php-debug/pull/293

src/xdebugConnection.ts
jonyo: @felixfbecker I agree, at the very least it should not emit something that only means something to vscode.

So, when you make any command that executes 1 or more lines of PHP code, the xdebug connection will not respond until the code is done executing, it is essentially locked up.  Given that, maybe it does make sense to have the logic in `xdebugConnection`?  Just change the name to be more appropriate like `code-execution-started`.  Instead of emmitting when vscode status should change to `ContinuedEvent` it is just emmitting that some command was called that will result in PHP code being executed.

I'm fine with doing it either way, changing the event name OR changing the caller to emit the `ContinuedEvent`.

Also if you think this requires more discussion I can move that change to it's own PR?  Since it is not required for the actual deadlock fix, it's more of a UX improvement, I can break it out if you want.

src/phpDebug.ts
jonyo: @felixfbecker Yeah, a more stable fix would be to somehow account for success, failed, or blocked and the blocked would not stop `configurationDone` from firing.  Ideally if there was some concrete way to tell that the breakpoint request was part of the process of initializing a new connection, I think it is still a good idea to only run it on the new connection in that case.

The tricky part here is that as far as I can tell these are initiated from vscode itself.  I used the word `likely` because of that, since _pendingConnections is not empty that means a new connection was just added.  For it to "not" be because of the new connection, the user would have had to do something like change a breakpoint at the precise moment a new connection was established, which is always a possibility.

Maybe we first: when doing something that executes code, set a flag like `_isLocked` that gets un-set when the response resolves.  Then when making the request to do something with breakpoints, if `connection._isLocked` it would queue the command but if there are more than one connection, as long as one succeeds it would continue.  That would allow the after configuration stuff to run for the new connection.

I hope that makes sense?

jonyo: In other words, instead of `Promise.All`, use something like `Promise.Any`, is that even a thing?

jonyo:
@felixfbecker Still getting errors.  I noticed there is now php_errors.log I wonder if it is something to do with php.ini settings?

```bash
  PHP Debug Adapter
    initialization
      ✓ should return supported features
    launch as CLI
      ✓ should error on non-existing file
      1) should run program to the end
      2) should stop on entry
      ✓ should not stop if launched without debugging
    continuation commands
      - should handle run
      - should handle step_over
      - should handle step_in
      - should handle step_out
      ✓ should error on pause request
      3) should handle disconnect
    breakpoints
      line breakpoints
        4) should stop on a breakpoint
        5) should stop on a breakpoint in file with spaces in its name
        6) should stop on a breakpoint identical to the entrypoint
        7) should support removing a breakpoint
      exception breakpoints
        8) should not break on anything if the file matches the ignore pattern
        9) should support stopping only on a notice
        10) should support stopping only on a warning
        - should support stopping only on an error
        11) should support stopping only on an exception
        12) should support stopping on everything
        - should report the error in a virtual error scope
      conditional breakpoints
        13) should stop on a conditional breakpoint when condition is true
        14) should not stop on a conditional breakpoint when condition is false
      function breakpoints
        15) should stop on a function breakpoint
    variables
      16) "before each" hook for "should report scopes correctly"
    virtual sources
      - should break on an exception inside eval code
      - should return the eval code with a source request
    parallel requests
      - should report multiple requests as threads
    evaluation
      - should return the eval result
      - should return variable references for structured results
    output events
      - stdout and stderr events should be complete and in correct order

  DbgpConnection
    ✓ should parse a response in one data event
    ✓ should parse a response over multiple data events
    ✓ should parse multiple responses in one data event
    ✓ should error on invalid XML

  paths
    isSameUri
      ✓ should compare to URIs
      ✓ should compare windows paths case-insensitive
    convertClientPathToDebugger
      without source mapping
        ✓ should convert a windows path to a URI
        ✓ should convert a unix path to a URI
      with source mapping
        ✓ should convert a unix path to a unix URI
        ✓ should convert a unix path to a windows URI
        - should convert a windows path to a unix URI
        - should convert a windows path with inconsistent casing to a unix URI
        - should convert a windows path to a windows URI
    convertDebuggerPathToClient
      without source mapping
        - should convert a windows URI to a windows path
        ✓ should convert a unix URI to a unix path
        - should handle non-unicode special characters
      with source mapping
        ✓ should map unix uris to unix paths
        - should map unix uris to windows paths
        ✓ should map windows uris to unix paths
        - should map windows uris to windows paths


  17 passing (13m)
  19 pending
  16 failing

  1) PHP Debug Adapter
       launch as CLI
         should run program to the end:
     Error: no event 'initialized' received after 10000 ms
      at Timeout.setTimeout [as _onTimeout] (node_modules/vscode-debugadapter-testsupport/lib/debugClient.js:222:28)

  2) PHP Debug Adapter
       launch as CLI
         should stop on entry:
     Error: no event 'stopped' received after 10000 ms
      at Timeout.setTimeout [as _onTimeout] (node_modules/vscode-debugadapter-testsupport/lib/debugClient.js:222:28)

  3) PHP Debug Adapter
       continuation commands
         should handle disconnect:
     Error: no event 'initialized' received after 10000 ms
      at Timeout.setTimeout [as _onTimeout] (node_modules/vscode-debugadapter-testsupport/lib/debugClient.js:222:28)

  4) PHP Debug Adapter
       breakpoints
         line breakpoints
           should stop on a breakpoint:
     Error: no event 'initialized' received after 10000 ms
      at Timeout.setTimeout [as _onTimeout] (node_modules/vscode-debugadapter-testsupport/lib/debugClient.js:222:28)

  5) PHP Debug Adapter
       breakpoints
         line breakpoints
           should stop on a breakpoint in file with spaces in its name:
     Error: no event 'initialized' received after 10000 ms
      at Timeout.setTimeout [as _onTimeout] (node_modules/vscode-debugadapter-testsupport/lib/debugClient.js:222:28)

  6) PHP Debug Adapter
       breakpoints
         line breakpoints
           should stop on a breakpoint identical to the entrypoint:
     Error: no event 'initialized' received after 10000 ms
      at Timeout.setTimeout [as _onTimeout] (node_modules/vscode-debugadapter-testsupport/lib/debugClient.js:222:28)

  7) PHP Debug Adapter
       breakpoints
         line breakpoints
           should support removing a breakpoint:
     Error: no event 'initialized' received after 10000 ms
      at Timeout.setTimeout [as _onTimeout] (node_modules/vscode-debugadapter-testsupport/lib/debugClient.js:222:28)

  8) PHP Debug Adapter
       breakpoints
         exception breakpoints
           should not break on anything if the file matches the ignore pattern:
     Error: no event 'initialized' received after 10000 ms
      at Timeout.setTimeout [as _onTimeout] (node_modules/vscode-debugadapter-testsupport/lib/debugClient.js:222:28)

  9) PHP Debug Adapter
       breakpoints
         exception breakpoints
           should support stopping only on a notice:
     Error: no event 'initialized' received after 10000 ms
      at Timeout.setTimeout [as _onTimeout] (node_modules/vscode-debugadapter-testsupport/lib/debugClient.js:222:28)

  10) PHP Debug Adapter
       breakpoints
         exception breakpoints
           should support stopping only on a warning:
     Error: no event 'initialized' received after 10000 ms
      at Timeout.setTimeout [as _onTimeout] (node_modules/vscode-debugadapter-testsupport/lib/debugClient.js:222:28)

  11) PHP Debug Adapter
       breakpoints
         exception breakpoints
           should support stopping only on an exception:
     Error: no event 'initialized' received after 10000 ms
      at Timeout.setTimeout [as _onTimeout] (node_modules/vscode-debugadapter-testsupport/lib/debugClient.js:222:28)

  12) PHP Debug Adapter
       breakpoints
         exception breakpoints
           should support stopping on everything:
     Error: no event 'initialized' received after 10000 ms
      at Timeout.setTimeout [as _onTimeout] (node_modules/vscode-debugadapter-testsupport/lib/debugClient.js:222:28)

  13) PHP Debug Adapter
       breakpoints
         conditional breakpoints
           should stop on a conditional breakpoint when condition is true:
     Error: no event 'initialized' received after 10000 ms
      at Timeout.setTimeout [as _onTimeout] (node_modules/vscode-debugadapter-testsupport/lib/debugClient.js:222:28)

  14) PHP Debug Adapter
       breakpoints
         conditional breakpoints
           should not stop on a conditional breakpoint when condition is false:
     Error: no event 'initialized' received after 10000 ms
      at Timeout.setTimeout [as _onTimeout] (node_modules/vscode-debugadapter-testsupport/lib/debugClient.js:222:28)

  15) PHP Debug Adapter
       breakpoints
         function breakpoints
           should stop on a function breakpoint:
     Error: no event 'initialized' received after 10000 ms
      at Timeout.setTimeout [as _onTimeout] (node_modules/vscode-debugadapter-testsupport/lib/debugClient.js:222:28)

  16) PHP Debug Adapter
       variables
         "before each" hook for "should report scopes correctly":
     Error: no event 'initialized' received after 10000 ms
      at Timeout.setTimeout [as _onTimeout] (node_modules/vscode-debugadapter-testsupport/lib/debugClient.js:222:28)



npm ERR! Test failed.  See above for more details.
```

felixfbecker:
Is XDebug enabled? What is `php -i`?

jonyo:
@felixfbecker 
[phpi.txt](https://github.com/felixfbecker/vscode-php-debug/files/2288699/phpi.txt)

felixfbecker:
You need to configure XDebug as described in the README, for example `xdebug.remote_autostart` is off in your info

jonyo:
Closing to restructure how it avoids the deadlock to be more stable per comments.

weinand:
@felixfbecker you asked:

> any advice how to handle this scenario best?

Sorry, but from the information in this issue I couldn't figure out what the scenario is.

jonyo:
@weinand By the way, here is the new PR:  #294 I made a new PR since it ended up being almost complete re-write, though in hindsight I should have just kept the original PR to make discussion easier.

Anyways, to answer what the scenario is:

1. request 1 makes a curl connection to a second endpoint.
2. At this point, xdebug fires up a 2nd connection.
3. vscode sends the signal to set up breakpoints.  But with how vscode works, it is not specific for a certain connection.
4. The xdebug extension tries to send the commands to get the current breakpoints, remove those, and add new ones.  On all connections.  And here is the breaking part: it waits to continue until the response resolves.
5. When it tries to do that for the 1st connection, those end up going on the command queue.
6. It just sits there, because connection 1 is waiting for the curl connection (connection 2) to finish.  Connection 2 is waiting for all the connections to be finished setting up breakpoints.  It will not send the "run" command to connection 2 until that happens.

This is potentially something that at the very least, causes the extension to hang, in other situations as well.  So in #294 I address that by introducing "execution commands", so that if the connection has issued a command resulting in PHP code being executed, the extension knows not to wait around for the response because it depends on how long PHP happens to take to get to the next stop point.  And in the above scenario it causes a deadlock (at least until curl times out).

PR #295, opened 2018-08-15
fix(phpDebug): Set evaluateName for variables to allow copy

### LINKED ISSUE #278: Unable to Copy/Set Variables within the debug Variables window
I've ask in other VSCode community site and all pointed to support within the debugger. As I'm using this PHP debugger it brought me here. My understanding is in other languages this works as expected. But for some reason in this PHP debugger when I right-click on a variables in the Variables' debug window I get the little popup showing options to Set Values, Copy Value, Copy Expression and Add to Watch but they are all greyed out and not functional. 

https://monosnap.com/file/1SQ78tzPQvlviGJ2pXFFxzvu1xjCry

Once I add items to the debug Watch window I can Copy, Set etc. But would really like to have this work under the Variables window .

My details:

Operating system: OS X 10.13.5
VSCode: 1.24.1
PHP Debugger: 1.12.3

launch.json file
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Listen for XDebug",
            "type": "php",
            "request": "launch",
            "port": 9000
        },
        {
            "type": "php",
            "request": "launch",
            "program": "${file}",
            "cwd": "${fileDirname}",
            "port": 9000
        }
    ]
}

jonyo:
Fixes #278 

This sets `evaluateName` on the variable object passed back to vscode.  I guess this is something from a newer version of vscode because I remember the copy used to work.

It sets it to `fullName` for the property if available (which will be something like `$var->property` if it is for a child), or `name` otherwise, which will be for top level variables.

### DIFF
--- a/src/phpDebug.ts
+++ b/src/phpDebug.ts
@@ -835,6 +835,7 @@ class PhpDebugSession extends vscode.DebugSession {
                 variables = properties.map(property => {
                     const displayValue = formatPropertyValue(property)
                     let variablesReference: number
+                    let evaluateName: string
                     if (property.hasChildren || property.type === 'array' || property.type === 'object') {
                         // if the property has children, we have to send a variableReference back to VS Code
                         // so it can receive the child elements in another request.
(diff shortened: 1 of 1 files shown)


jonyo:
Note: I found it thanks to one of the comments on #278 that pointed to Microsoft/ptvsd#559 - the PR there gave the hint that what was missing, `evaluateName` needed to be set.

PR #307, opened 2018-09-19
fix: do not send continue events

### LINKED ISSUE #301: Debug context lost on multiple connections
Relates to: #19 #167 #294 #231
PHP version:        7.2.7
XDebug version:  2.6.0
Adapter version:  1.12.5

# Issue
When multiple connections are received, the debug context of the previous connection(s) cannot be retrieved from the editor.

For larger codebases (tested with WordPress) it seems that both debug contexts are lost resulting in an empty 'Call Stack', meaning the 'Call Stack' window is empty. The first connection is kept paused, whereas the second gets executed without pausing.

# Steps to reproduce
Given the following PHP script `debug-multi.php`:
```php
$a = ['bee', 'tiger', 'bear', 'deer'];
xdebug_break();
var_dump($a);
```

**Step 1**
Open a terminal and execute `export XDEBUG_CONFIG="idekey=vscode"; php debug-multi.php`.
The adapter correctly stops on the line following `xdebug_break()`:

image

image

**Step 2**
Open a terminal and execute `export XDEBUG_CONFIG="idekey=vscode"; php debug-multi.php`.
The adapter correctly stops on the same line as step 1, but **the previous request is shown as "Running"** even if **it is still paused** (shell command has not returned):

image

image

# Configuration settings
launch.json:
```
"name": "Listen for XDebug",
"type": "php",
"request": "launch",
   "port": 9000,
   "xdebugSettings": {
   "max_depth": 1,
    "max_children": 10
},
"log": true
```
php.ini:
```
xdebug.remote_enable = 1
xdebug.profiler_output_dir=/Users/ettore/Workspace/xdebug/profiling
xdebug.profiler_output_name=%u.%R
xdebug.profiler_enable_trigger=1
xdebug.remote_log = /Users/ettore/Workspace/xdebug/xdebug.log
```
# Logs

## First connection

[1-adapter.log](https://github.com/felixfbecker/vscode-php-debug/files/2348342/1-adapter.log)

## Second connection

[2-adapter.log](https://github.com/felixfbecker/vscode-php-debug/files/2348344/2-adapter.log)

jonyo:
…f the normal flow

This fixes #301 

Note: there is still the problem that it will seem to "hang" when you continue and the PHP code takes a while to run, #308 will fix that.

### DIFF
--- a/src/phpDebug.ts
+++ b/src/phpDebug.ts
@@ -279,10 +279,6 @@ class PhpDebugSession extends vscode.DebugSession {
                         })
                         connection.on('error', disposeConnection)
                         connection.on('close', disposeConnection)
-                        connection.on('before-execute-command', () => {
-                            // It is about to start executing PHP code
-                            this.sendEvent(new vscode.ContinuedEvent(connection.id))
-                        })
                         await connection.waitForInitPacket()
 
                         // override features from launch.json
--- a/src/xdebugConnection.ts
+++ b/src/xdebugConnection.ts
@@ -679,11 +679,6 @@ export class Connection extends DbgpConnection {
         const data = iconv.encode(commandString, ENCODING)
         this._pendingCommands.set(transactionId, command)
         this._pendingExecuteCommand = command.isExecuteCommand
-        if (this._pendingExecuteCommand) {
-            // Since PHP execution commands block anything on the connection until it is
-            // done executing, emit that the connection is about to go into such a locked state
-            this.emit('before-execute-command')
-        }
         await this.write(data)
     }
 


src/phpDebug.ts
```diff
@@ -279,10 +279,6 @@ class PhpDebugSession extends vscode.DebugSession {
                         })
                         connection.on('error', disposeConnection)
                         connection.on('close', disposeConnection)
-                        connection.on('before-execute-command', () => {
```
felixfbecker: Is this event then still used at all?

src/xdebugConnection.ts
```diff
@@ -679,11 +679,6 @@ export class Connection extends DbgpConnection {
         const data = iconv.encode(commandString, ENCODING)
         this._pendingCommands.set(transactionId, command)
         this._pendingExecuteCommand = command.isExecuteCommand
-        if (this._pendingExecuteCommand) {
```
felixfbecker: And this flag?

src/phpDebug.ts
jonyo: @felixfbecker No, this PR completely removes the event, it turns out that it causes problems to send the continued event.  In several scenarios it duplicates other events sent that also tell vscode that the code is running.  The second notice sometimes gets sent after it is paused again, and ends up making vsc think it continued when it is actually still paused.

src/xdebugConnection.ts
jonyo: @felixfbecker The flag itself is still used to prevent that deadlock where it waits for breakpoint init call on a connection that is hung running code.

jonyo:
@felixfbecker It helps to see the description for that event:

From https://github.com/Microsoft/vscode-debugadapter-node/blob/f479e3dc97796cecaebccf455707cd4b5a4c0ad1/protocol/src/debugProtocol.ts#L107-L111

```php
	/** Event message for 'continued' event type.
		The event indicates that the execution of the debuggee has continued.
		Please note: a debug adapter is not expected to send this event in response to a request that implies that execution continues, e.g. 'launch' or 'continue'.
		It is only necessary to send a 'continued' event if there was no previous request that implied this.
	*/
```

So in other words, if there is somewhere that we are making the code continue that is not part of a continue request from vscode, that is when to use it.  One thing that comes to mind is if the code has paused somewhere that we ignore and we want to continue automatically, we might want to then notify vscode that it has continued.

My original statement was before I fully looked into it.  It would be more accurate to say that we send the continued event *in addition to* the normal response to vscode, and sometimes that can cause vscode to think the code is running when it is actually paused.  One example of when this could happen: it sends as a result of initializing xdebug as part of the first run command.  It sends that continued event. That event is sent async.  It could actually get to a breakpoint and be paused on the breakpoint at the time vscode receives the event, so vscode incorrectly thinks the code has continued when it has not.  That is what was happening on #301 I hope that makes sense?

felixfbecker:
Ah, that makes sense to me.

PR #308, opened 2018-09-19
Refactor continue events to avoid UI hanging

jonyo:
If you click next or continue and the PHP code is taking a while to run, it seems like the button did not have an effect because the context still shows and it still appears to be paused.  The only hint that it worked, is that if you try to interact with the context, or eval any code in the console, it won't work because it has actually continued.

Looking at the docs for the continue, step into, etc. events, they indicate that it should actually send the response right away indicating that the action has started.  Then after the action is complete, send the appropriate stop event depending on if it is now stopped on breakpoint etc.

But the way it was before, it did not do this, it waited until the next pause point before sending both the initial response and the one run after it is done.  This was causing the UI to hang as described above.

Here is the applicable reference documentation I relied on when making this change:

https://github.com/Microsoft/vscode-debugadapter-node/blob/750f50bc875669fe3cb7594804bcfc6326b27d0e/protocol/src/debugProtocol.ts#L559

> The debug adapter **first** sends the response **and then** a 'stopped' event (with reason 'step') **after the step has completed**.

(my emphasis added)

Relates to #307 #301

### DIFF
--- a/src/phpDebug.ts
+++ b/src/phpDebug.ts
@@ -945,7 +945,6 @@ class PhpDebugSession extends vscode.DebugSession {
         response: VSCodeDebugProtocol.DisconnectResponse,
         args: VSCodeDebugProtocol.DisconnectArguments
     ) {
-        this.sendResponse(response)
         try {
             await Promise.all(
                 Array.from(this._connections).map(async ([id, connection]) => {
@@ -969,6 +968,7 @@ class PhpDebugSession extends vscode.DebugSession {
             this.sendErrorResponse(response, error)
             return
         }
+        this.sendResponse(response)
         this.shutdown()
     }
 
--- a/src/phpDebug.ts
+++ b/src/phpDebug.ts
@@ -889,88 +889,52 @@ class PhpDebugSession extends vscode.DebugSession {
         response: VSCodeDebugProtocol.ContinueResponse,
         args: VSCodeDebugProtocol.ContinueArguments
     ) {
-        let xdebugResponse: xdebug.StatusResponse | undefined
-        try {
-            const connection = this._connections.get(args.threadId)
-            if (!connection) {
-                throw new Error('Unknown thread ID ' + args.threadId)
-            }
-            xdebugResponse = await connection.sendRunCommand()
-        } catch (error) {
-            this.sendErrorResponse(response, error)
-            if (xdebugResponse) {
-                this._checkStatus(xdebugResponse)
-            }
-            return
-        }
         response.body = {
             allThreadsContinued: false,
         }
-        this.sendResponse(response)
-        this._checkStatus(xdebugResponse)
-    }
-
-    protected async nextRequest(response: VSCodeDebugProtocol.NextResponse, args: VSCodeDebugProtocol.NextArguments) {
-        let xdebugResponse: xdebug.StatusResponse | undefined
-        try {
-            const connection = this._connections.get(args.threadId)
-            if (!connection) {
-                throw new Error('Unknown thread ID ' + args.threadId)
-            }
-            xdebugResponse = await connection.sendStepOverCommand()
-        } catch (error) {
-            this.sendErrorResponse(response, error)
-            if (xdebugResponse) {
-                this._checkStatus(xdebugResponse)
-            }
+        const connection = this._connections.get(args.threadId)
+        if (!connection) {
+            this.sendErrorResponse(response, new Error('Unkown thread ID ' + args.threadId))
             return
         }
         this.sendResponse(response)
-        this._checkStatus(xdebugResponse)
+        this._checkStatus(await connection.sendRunCommand())
+    }
+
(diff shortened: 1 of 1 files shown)


src/phpDebug.ts
```diff
@@ -279,10 +279,6 @@ class PhpDebugSession extends vscode.DebugSession {
                         })
                         connection.on('error', disposeConnection)
                         connection.on('close', disposeConnection)
-                        connection.on('before-execute-command', () => {
```
felixfbecker: Is that event still used at all?

src/phpDebug.ts
```diff
@@ -893,88 +889,52 @@ class PhpDebugSession extends vscode.DebugSession {
...
             allThreadsContinued: false,
         }
+        const connection = this._connections.get(args.threadId)
+        if (!connection) {
+            this.sendErrorResponse(response, new Error('Unkown thread ID ' + args.threadId))
+            return
+        }
         this.sendResponse(response)
-        this._checkStatus(xdebugResponse)
+        this._checkStatus(await connection.sendRunCommand())
```
felixfbecker: The problem with this is that you are not handling the error anymore. The response to this command is the only chance to report that the run command failed. XDebug doesn't return a response to the `continue` command unless the command failed or the program stopped again.

If the only way is to send the response straight away, it needs to at least log errors, or only send the response after a timeout.

src/phpDebug.ts
```diff
@@ -1008,7 +969,6 @@ class PhpDebugSession extends vscode.DebugSession {
             this.sendErrorResponse(response, error)
```
felixfbecker: This line is not valid anymore because the response has already been sent.

src/phpDebug.ts
jonyo: @felixfbecker 

> If the only way is to send the response straight away, it needs to at least log errors, or only send the response after a timeout.

The docs say to send the response that says the action was started, so yes it needs to send the response right away.  As I understand it, it should use `sendErrorResponse()` only if there is some reason it cannot start the action.  The results of the action are sent up the line to vscode in other ways already, these are the different outcomes I can think of:

* success, now it is paused again - we let xdebug know in `_checkStatus()`
* Some exception / error like fatal error - this is also sort of success, the code was run and it is now paused on some problem in the code, so we let vscode know about it in `_checkStatus()`
* The script finishes.  We let vscode know about that on https://github.com/felixfbecker/vscode-php-debug/blob/cf7ad01339f5ea053cc5117f101d6db085112492/src/phpDebug.ts#L281
* There is some connection level error, we let vscode know about those here: 
 https://github.com/felixfbecker/vscode-php-debug/blob/cf7ad01339f5ea053cc5117f101d6db085112492/src/phpDebug.ts#L277-L280
* There is a server level error, we let vscode know about those here:  https://github.com/felixfbecker/vscode-php-debug/blob/cf7ad01339f5ea053cc5117f101d6db085112492/src/phpDebug.ts#L309-L312

It looked like the try/catch was specifically for handling when the connection was not found with that `threadId`, and when that is the case, we now handle it directly by responding with `sendErrorResponse()`.  As far as I could tell it would never throw an exception as part of the call to `connection.sendRunCommand()`, instead it would either return a response with an error in it (handled by `_checkStatus()`, or emit an event which the `phpDebug` is already handling on it's own as noted above.

All vscode needs to know here is that we started running the command, after that point if there is an error we let vscode know about it in other emitted events.

src/phpDebug.ts
jonyo: @felixfbecker I ended up reverting this part to how it was.  Really the docs are rather vague so we are forced to figure out whether we should send immediately or wait.  My original thought was to not wait since the response has no information to tell vscode, it is just an acknowledgment.

But now I'm thinking, this is a disconnect response, there is no harm in waiting, especially since it already has timeouts built in.

In addition, this could be a part of a restart request, in which case we do definitely want to wait before telling vscode, I could see a situation where we let vscode know too early and it results in vscode attempting to start a new connection before the old one is done.  You would end up with errors about the port already being in use.

**tl;dr:** No harm in making it work like it did before, and there is potentially a problem by not waiting, so better to be on the safe side since we don't have much direction from the docs.

src/phpDebug.ts
felixfbecker: Sorry for the late reply. I agree with what you say, but that doesn't change the fact that `sendRunCommand()` can reject with an exception, and that exception needs to be handled _somehow_. Even if it is handled async and we send the response before, it needs an error handler that at a minimum needs to send an error log to VS Code.

jonyo: Ah, I see what you mean, if the xdebug socket is not writable it will reject so need to catch that.

If you are in agreement about using async calls (with addition of handling the reject error) let me know.  If interested I'd be happy to rebase (if possible after a year) or just redo this PR when I have some time.

felixfbecker: Yes that sounds good

zobo:
Already solved by #367.

PR #312, opened 2018-10-03
fix XDebug installation steps list

odahcam:


felixfbecker:
Whats fixed here?

odahcam:
If you look at master, you can see that item 3 breaks the list, its contents closes the `ol` and for item 4 a new `ol` is rendered. I fixed this break by correctly identating the content, now generating only one `ol`.

felixfbecker:
CI is failing but passing on master

odahcam:
Seems like a linting problem pointed by Prettier, I formated the code following markdown good practices so it must be something with line length from Prettier. :/

I'll clone the repo later to be able to run the linter and fix this.

PR #315, opened 2018-10-12
ci: Ability to open a browser when the debug session is created

letynsoft:
Adds launch option which allows setting the command to run and the URL to open (to open a browser) with the xdebug query param to start the debugging session


letynsoft:
Hi, i will implement the other things you menthined, but when i try to add the open module it just wont create the packages-lock.jsosn file (which seems to be neerded for the checking stuff), can you help me with that (tried simple npm install --save open, but as i say, doesn't update the lock file)?
Also not sure what you mean with the camel case one, but i guess, once i figure out the open package, it will only need a string URL there

felixfbecker:
Are you running at least node 8 and npm 6?

letynsoft:
Hi, well i try to keep it up-to-date:
# npm version
{ npm: '6.4.1',
  ares: '1.14.0',
  cldr: '32.0.1',
  http_parser: '2.7.1',
  icu: '60.2',
  modules: '57',
  nghttp2: '1.30.0',
  node: '8.10.0',
  openssl: '1.1.0g',
  tz: '2017c',
  unicode: '10.0',
  uv: '1.18.0',
  v8: '6.2.414.50',
  zlib: '1.2.11' }

letynsoft:
And of course it does update it now... :-/ ... ok will do the other things now... so don't bother :)

letynsoft:
Hi,
So i have updated the change.
Let me know your thoughts :)

src/phpDebug.ts
```diff
@@ -12,6 +12,9 @@ import { Terminal } from './terminal'
 import { isSameUri, convertClientPathToDebugger, convertDebuggerPathToClient } from './paths'
 import minimatch = require('minimatch')
 
+let open = require('open')
```
felixfbecker: Please use `import open from 'open'` or this is not type safe. You'll need to add `npm install --save-dev @types/open`

src/phpDebug.ts
```diff
@@ -12,6 +12,9 @@ import { Terminal } from './terminal'
...
+let appendQuery = require('append-query')
```
felixfbecker: This is built into Node, no need for a dependency. Just use `import { URL } from 'url'`

src/phpDebug.ts
```diff
@@ -311,6 +317,12 @@ class PhpDebugSession extends vscode.DebugSession {
                     this.sendErrorResponse(response, <Error>error)
                 })
                 server.listen(args.port || 9000, (error: NodeJS.ErrnoException) => (error ? reject(error) : resolve()))
+                if (args.openUrl) {
+                    let proc = open(appendQuery(args.openUrl, 'XDEBUG_SESSION_START=vscode'))
+                    proc.stderr.on('data', (data: any) => {
```
felixfbecker: `data` is not `any`, it's either a string or a Buffer, in which case the logged output would be useless (octets). I don't know from the top of my head what it is, but it should not be typed `any`.

src/phpDebug.ts
```diff
@@ -311,6 +317,12 @@ class PhpDebugSession extends vscode.DebugSession {
...
+                        this.sendEvent(new vscode.OutputEvent('openUrl: ' + util.inspect(data) + '\n'))
```
felixfbecker: You shouldn't use `inspect` because it's not a structured object

src/phpDebug.ts
```diff
@@ -992,6 +1004,12 @@ class PhpDebugSession extends vscode.DebugSession {
                     this._waitingConnections.delete(connection)
                 })
             )
+            if (this._args.openUrl) {
+                let proc = open(appendQuery(this._args.openUrl, 'XDEBUG_SESSION_STOP=vscode'))
+                proc.stderr.on('data', (data: any) => {
+                    this.sendEvent(new vscode.OutputEvent('openUrl: ' + util.inspect(data) + '\n'))
+                })
+            }
```
felixfbecker: This code is duplicated, it should be in a shared function

src/phpDebug.ts
```diff
@@ -311,6 +317,12 @@ class PhpDebugSession extends vscode.DebugSession {
                     this.sendErrorResponse(response, <Error>error)
                 })
                 server.listen(args.port || 9000, (error: NodeJS.ErrnoException) => (error ? reject(error) : resolve()))
+                if (args.openUrl) {
+                    let proc = open(appendQuery(args.openUrl, 'XDEBUG_SESSION_START=vscode'))
```
felixfbecker: Use `const`

letynsoft:
Hi,
Another update...
Seems i didn't notice the first open should have been opn in the commit title :-/

letynsoft:
Any news here? :)

package.json
```diff
@@ -34,10 +34,12 @@
     "url": "https://github.com/felixfbecker/vscode-php-debug/issues"
   },
   "dependencies": {
+    "append-query": "^2.0.1",
```
felixfbecker: This dependency is now unused

src/phpDebug.ts
```diff
@@ -325,6 +330,30 @@ class PhpDebugSession extends vscode.DebugSession {
...
+        }
+
+        const proc = opn(myurl.toString())
+        proc.then(child => {
+            child.stderr.on('data', (chunk: string) => {
+                this.sendEvent(new vscode.OutputEvent('openUrl: ' + chunk + '\n'))
+            })
+        }).catch((reason: any) => {
+            this.sendEvent(new vscode.OutputEvent('openUrl-error: ' + util.inspect(reason) + '\n'))
+        })
```
felixfbecker: Instead of loggint the error here, it’s better to make this an async function and have the callers await it, so an error is reported as a response to the request

src/phpDebug.ts
```diff
@@ -325,6 +330,30 @@ class PhpDebugSession extends vscode.DebugSession {
...
+     * Opens the urls specified in configuration, either with start or stop query string
+     * @param {boolean} start
+     */
+    private _openUrl(start: boolean) {
+        if (!this._args.openUrl) {
+            return
+        }
+        const myurl = new url.URL(this._args.openUrl)
+        if (start) {
+            myurl.searchParams.append('XDEBUG_SESSION_START', 'vscode')
```
felixfbecker: I think this should be set, not append, since append would duplicate the parameter if it’s already set

letynsoft: I would rather keep the append, as a the stop command would be confusing for the xdebug anyway

src/phpDebug.ts
letynsoft: Not sure what exactly you mean here... so i should remove the .catch and just return the result?
Also should the then be removed too?

src/phpDebug.ts
felixfbecker: I don't understand what this has to do with the stop command - what I mean is that `append` would potentially add `XDEBUG_SESSION_START` twice or more times, like `?XDEBUG_SESSION_START=vscode&XDEBUG_SESSION_START=vscode`, while `set()` would just override it so `XDEBUG_SESSION_START` appears exactly once

src/phpDebug.ts
```diff
@@ -325,6 +330,30 @@ class PhpDebugSession extends vscode.DebugSession {
         }
         this.sendResponse(response)
     }
+    /**
+     * Opens the urls specified in configuration, either with start or stop query string
+     * @param {boolean} start
+     */
+    private _openUrl(start: boolean) {
```
felixfbecker:
```suggestion
    private async _openUrl(start: boolean): Promise<void> {
```

src/phpDebug.ts
```diff
@@ -325,6 +330,30 @@ class PhpDebugSession extends vscode.DebugSession {
...
+            return
+        }
+        const myurl = new url.URL(this._args.openUrl)
+        if (start) {
+            myurl.searchParams.append('XDEBUG_SESSION_START', 'vscode')
+        } else {
+            myurl.searchParams.append('XDEBUG_SESSION_STOP', 'vscode')
+        }
+
+        const proc = opn(myurl.toString())
```
felixfbecker:
```suggestion
        const proc = await opn(myurl.toString())
```

src/phpDebug.ts
felixfbecker: See my two suggestions above. In addition you'd remove the `.catch()` and "unwrap" the `.then()`.

src/phpDebug.ts
letynsoft: but what would happen when the stop command should be sent? XDEBUG_SESSION_START=vscode&XDEBUG_SESSION_STOP=vscode (with set or append) as the param is diferrent, so it would be easier to spot such errors.

ianfoulds:
Hi @letynsoft  @felixfbecker
This is an excellent step forward. I've just tried to implement a simple solution using the "program" and "args" properties but it stops debugging when the "program" script ends.

Any idea on when you might get this completed, it's a feature I would really like to see. I've just moved back to VSC from NetBeans which had a pretty good config for doing this and I'm missing it already.

As an aside, could you also add something along the lines of "program" (very weird choice of parameter name for script) and "args" that does not terminate the debug session which it finishes? Maybe a "init" object with "program" or "script" and "args" that are run when the debugger is started. This would allow platform specific code to run via exec which could open a browser (start http://.... on Windows) or all manner of other funky stuff.

zobo:
I think implementing this in this way just keeps adding complexity to the DA for things that it should not be directly responsible for. I think that very similar can be achieved with either [`preLaunchTask`](https://code.visualstudio.com/docs/debugtest/debugging-configuration#_launchjson-attributes) or [`serverReadyAction`](https://code.visualstudio.com/docs/debugtest/debugging-configuration#_automatically-open-a-uri-when-debugging-a-server-program). If people still want this, I'd rather implement it in the extension code (not DA) with commands or tasks.

Issue #346, opened 2019-04-10
Extension not exporting activate function


tsmaeder:
The [VSCode Documentation](https://code.visualstudio.com/api/references/activation-events#Start-up) requires an extension to export an activate function. However, vscode-php-debug does not seem to export one. While VS Code happily load the extension currently, this might change at any time.

zobo:
This was a DAP-only extension and so did not need an `activate()` export. It was changed and expanded a lot since then...

PR #347, opened 2019-04-10
fix: Extension not exporting activate function (#346)

vrubezhny:
The fix adds `main` attribute to `package.json` thus making it to be treated an an extension.

Signed-off-by: Victor Rubezhny <vrubezhny@redhat.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
   "bugs": {
     "url": "https://github.com/felixfbecker/vscode-php-debug/issues"
   },
+  "main": "./out/extension",
   "dependencies": {
     "file-url": "^2.0.0",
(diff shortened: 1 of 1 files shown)


RobberPhex:
You just point the `main` to a non-existing file. I don't think it would work.

zobo:
This is currently not needed because no extension functionality is present in this code, only the Debug Adapter. To quote from docs:

> So in its most minimal form, a debugger extension is just a declarative contribution of a debug adapter implementation and the extension is basically a packaging container for the debug adapter without any additional code.

See https://code.visualstudio.com/api/extension-guides/debugger-extension#debugging-architecture-of-vs-code

PR #362, opened 2019-06-12
feat: launch support auto port

RobberPhex:
If user doesn't specify a debug port,
The script will use a free port to use.

Sometimes, user just want debug a script, and doesn't care about debug port. so, let it choose automatically.

### DIFF
--- a/package.json
+++ b/package.json
@@ -250,7 +250,7 @@
             "request": "launch",
             "program": "${file}",
             "cwd": "${fileDirname}",
-            "port": 9000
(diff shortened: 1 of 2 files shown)


felixfbecker:
Could you give a description of the change in the PR description?

RobberPhex:
@felixfbecker I've add description, if you have any problem, please comment here.

zobo:
This was implemented with PR #588 and others (see #575).

PR #363, opened 2019-06-14
feat: implement longest prefix matching for pathMappings (closes #331)

### LINKED ISSUE #331: Longest prefix matching for pathMapping server paths?
In my docker container with XDebug I have two volumes mounted: 

- `/var/www/html` mounts a `wordpress` folder in my Code workspace. That folder is ignored by git, as it only contains a WordPress installation.
- `/var/www/html/wp-content/themes/my-theme` mounts a  `my-theme` folder in my workspace, which is subject of my work and tracked by git.

I have added pathMappings to my launch.json like this:
```json
"pathMappings": {
     "/var/www/html/wp-content/themes/my-theme": "${workspaceRoot}/my-theme",
     "/var/www/html": "${workspaceRoot}/wordpress"
}
```

Unfortunately, as soon as I include `"/var/www/html"` in the `pathMappings`, server files within `/var/www/html/wp-content/themes/my-theme` are mapped to the corresponding `${workspaceRoot}/wordpress` subdirectory, instead of the `${workspaceRoot}/my-theme` folder.
What I would like the extension to do in this scenario is performing a longest prefix matching for server paths.
Are there any plans for implementing this in the future? Or am I missing a point here maybe?
If this feature is welcome, I would certainly offer to implement it!

bjoluc:
I decided to approach #331 and it turned out to be only a minor change. I also added longest prefix matching for the client to debugger mapping as there are use cases for this as well and it does not affect the existing mapping behavior (unless in actual use cases in which the change is most certainly welcome).

### DIFF
--- a/src/test/paths.ts
+++ b/src/test/paths.ts
@@ -55,6 +55,15 @@ describe('paths', () => {
                     }),
                     'file:///app/source.php'
                 )
+                // longest prefix matching for server paths
+                assert.strictEqual(
+                    convertClientPathToDebugger('/home/felix/mysource/subdir/source.php', {
+                        '/var/www': '/home/felix/mysite',
+                        '/app/subdir1': '/home/felix/mysource/subdir',
+                        '/app': '/home/felix/mysource',
+                    }),
+                    'file:///app/subdir1/source.php'
+                )
             })
             // unix to windows
             it('should convert a unix path to a windows URI', () => {
@@ -164,6 +173,15 @@ describe('paths', () => {
                         '/app': '/home/felix/mysource',
                     }),
                     '/home/felix/mysource/source.php'
+                )
+                // longest prefix matching for local paths
+                assert.strictEqual(
+                    convertDebuggerPathToClient('file:///app/subdir/source.php', {
+                        '/var/www': '/home/felix/mysite',
+                        '/app/subdir': '/home/felix/mysource/subdir1',
+                        '/app': '/home/felix/mysource',
+                    }),
+                    '/home/felix/mysource/subdir1/source.php'
                 )
             })
             // unix to windows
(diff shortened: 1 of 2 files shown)


bjoluc:
Hm, the Travis builds fail to install Xdebug. @felixfbecker Do you have any idea why?

killua99:
Activating this.

Shall we take a look if this could be implemented ? this case scenario is quite common for WordPress development often wp-content is on its own and the WordPress folder maps differently.

zobo:
Hi @bjoluc . Can you rebase this onto main? I'll have a fresh look at it then.

bjoluc:
Sure, thanks!

src/test/paths.ts
```diff
@@ -55,6 +55,15 @@ describe('paths', () => {
                     }),
                     'file:///app/source.php'
                 )
+                // longest prefix matching for server paths
+                assert.strictEqual(
+                    convertClientPathToDebugger('/home/felix/mysource/subdir/source.php', {
+                        '/var/www': '/home/felix/mysite',
+                        '/app': '/home/felix/mysource',
+                        '/app/subdir1': '/home/felix/mysource/subdir',
```
zobo, line 63: The list should be built so that without the patch the test fails. Please move `'/app/subdir1': '/home/felix/mysource/subdir',` one line up, before `'/app': '/home...`.

src/test/paths.ts
```diff
@@ -165,6 +174,15 @@ describe('paths', () => {
                     }),
                     '/home/felix/mysource/source.php'
                 )
+                // longest prefix matching for server paths
+                assert.strictEqual(
+                    convertDebuggerPathToClient('file:///app/subdir/source.php', {
+                        '/var/www': '/home/felix/mysite',
+                        '/app': '/home/felix/mysource',
+                        '/app/subdir': '/home/felix/mysource/subdir1',
```
zobo, line 182: The list should be built so that without the patch the test fails. Please move `'/app/subdir': '/home/felix/mysource/subdir1',` one line up, before `'/app': '/home...`.

zobo (changes requested):

src/test/paths.ts
bjoluc: Good catch!

zobo (approved):

zobo:
Looks great. I'll to another round of tests here and merge it then. Thanks!

PR #365, opened 2019-06-22
fix: convertDebuggerPathToClient

RobberPhex:
Win to unix, the multi-level path cannot convert correctly.

### DIFF
--- a/src/paths.ts
+++ b/src/paths.ts
@@ -40,10 +40,9 @@ export function convertDebuggerPathToClient(
             serverSourceRoot,
             serverPath
         )
-        pathRelativeToSourceRoot = pathRelativeToSourceRoot.replace(
-            (serverIsWindows ? path.win32 : path.posix).sep,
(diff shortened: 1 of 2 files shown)


src/paths.ts
```diff
@@ -36,10 +36,13 @@ export function convertDebuggerPathToClient(
     let localPath: string
     if (serverSourceRoot && localSourceRoot) {
         // get the part of the path that is relative to the source root
-        const pathRelativeToSourceRoot = (serverIsWindows ? path.win32 : path.posix).relative(
+        let pathRelativeToSourceRoot = (serverIsWindows ? path.win32 : path.posix).relative(
```
felixfbecker: So if `serverIsWindows`, first we explicitly use `path.win32`, which always outputs backslashes, and then we replace those again with `path.sep`?

zobo:
Incorporated in #525.

PR #366, opened 2019-07-08
Support terminateThreads

zobo:
Add support for terminate threads. Useful when more than one scripts are being debugged and we want to terminate only one instead of all.

Before this PR can be applied the vscode-debugadapter requires a fix submitted in PR https://github.com/microsoft/vscode-debugadapter-node/pull/214


zobo:
I suggest merging #378 first, as that one is bigger and includes package fixes already.

PR #367, opened 2019-07-15
fix: remove current execution position when performing stop or continue dbgp operation.

### LINKED ISSUE #358: Feature request: Remove stopped indicator
When you hit a break point and then resume the execution (F5), I think it would be best to remove the yellow arrow from the gutter and also the yellow highlight of the last executed line. This would indicate that the code is executing normally again. Currently, even if the execution was resumed, the yellow arrow and the yellow highlight stay, which is very confusing as you can never be sure if the execution was correctly resumed or not.

The same is true for the call stack view, which displays the "PAUSED ON ..." label even after resuming the execution. Basically there is no way to know if your code is running or not after resuming the execution.

Thanks!

zobo:
When the IDE requests a continue operation (step in, out, over, continue) it expects a Response and will remove the current execution position. With XDebug/DBGp the TCP connection will not send a response until another stop condition is encountered this may take an arbitrary amount of time.
This change fixes this by doing some basic validation (is the threadId valid) and then sends a response before executing the continuation command.

Closes #358

### DIFF
--- a/src/phpDebug.ts
+++ b/src/phpDebug.ts
@@ -894,9 +894,9 @@ class PhpDebugSession extends vscode.DebugSession {
         response: VSCodeDebugProtocol.ContinueResponse,
         args: VSCodeDebugProtocol.ContinueArguments
     ) {
-        let xdebugResponse: xdebug.StatusResponse | undefined
+        let connection: xdebug.Connection | undefined
         try {
-            const connection = this._connections.get(args.threadId)
+            connection = this._connections.get(args.threadId)
             if (!connection) {
                 return this.sendErrorResponse(response, new Error('Unknown thread ID ' + args.threadId))
             }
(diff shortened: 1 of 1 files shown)


zobo:
I rebased the code to latest main and fixed the title to pass all checks.
The PR also contains a different way to handle possible missing thread ID error. This way the error will, besides being shown as a toast in vscode, also be logged in the Debug console.

src/phpDebug.ts
```diff
@@ -898,20 +898,19 @@ class PhpDebugSession extends vscode.DebugSession {
...
-                throw new Error('Unknown thread ID ' + args.threadId)
+                return this.sendErrorResponse(response, new Error('Unknown thread ID ' + args.threadId))
             }
+            response.body = {
+                allThreadsContinued: false,
+            }
+            this.sendResponse(response)
             xdebugResponse = await connection.sendRunCommand()
         } catch (error) {
-            this.sendErrorResponse(response, error)
```
felixfbecker, line 905: If this is removed, it would be good to at least log the error to the user to the debug console.

zobo: I have refactored the code to handle all error scenarios. Those before sending a response and those after already sending a response.

I also written a generalized helper method, since all step calls (continueRequest, nextRequest, stepInRequest and stepOutRequest) are basically the same.
I have not committed it, as it does introduce some indirection and makes the code a bit harder to follow. Here's the function, we can still use it if you think it's worth it.

```typescript
    protected async stepRequestHelper<T extends VSCodeDebugProtocol.Response>(name: string, threadId: number, response: T, op: (connection: xdebug.Connection)=>Promise<xdebug.StatusResponse>) {
        let connection: xdebug.Connection | undefined
        try {
            connection = this._connections.get(threadId)
            if (!connection) {
                return this.sendErrorResponse(response, new Error('Unknown thread ID ' + threadId))
            }
            this.sendResponse(response)
        } catch (error) {
            return
        }
        try {
            const xdebugResponse = await op(connection)
            this._checkStatus(xdebugResponse)
        } catch (error) {
            this.sendEvent(new vscode.OutputEvent(`${name} thread Id ${threadId} error: ${error.message}\n`))
        }
    }
```

marciovennan:
This update caused a problem when running the debug. When I run the debug it is closed automatically. I have already uninstalled the plugin and also vscode, but the problem persists.

https://user-images.githubusercontent.com/3438044/106654586-130c0300-6577-11eb-8589-6da0452962ff.mp4

zobo:
Bug came in with previous commits. Was fixed in 1.14.7

PR #368, opened 2019-07-15
fix: removes code execution indicator when process is terminated

zobo:
If the debugger is standing on a line and DBGp connection is interrupted (for example, php process terminated) the thread will go away, but the execution indicator will stay. Signaling a continued event before Thread exit solves this.


elovin:
@felixfbecker any plans when this will be merged ?

PR #356, opened 2019-05-04
Fix case: XDebug -> Xdebug

angristan:
This is not a troll or anything, I was asking myself what's the correct way to write Xdebug and noticed that "XDebug" is mainly used is this extension. Is it intentional?

Source: https://xdebug.org/

PR #374, opened 2019-09-23
fix Xdebug author name

peterdd:
~~and at the moment some links to xdebug.org are broken, but @derickr will probably fix them now on xdebug.org I assume. (mostly links ending with .php work without the ending, prob server conf error)~~ (fixed now on xdebug.org)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 Install the extension: Press `F1`, type `ext install php-debug`.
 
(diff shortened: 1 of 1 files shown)


peterdd:
> The page also really should use Xdebug and not XDebug.
There is pull request #356 waiting since may 2019.

zobo:
Closed as it was solved in #476.

PR #280, opened 2018-06-25
Adding DBGp Proxy support

SidIcarus:
Related #107, #110 , #111, & #274 
successfully allows for multiple concurrent connections to a double nat'd network

- [x] Handle response data
- [x] tests
  - [x] errors: timeout, failure to resolve, duplicate key,  no key
  - [x] successful registration, de-registration 
- [x] ~proxy error command~

PR #378, opened 2019-11-15
Add DBGP proxy support, with client local port

zobo:
Changes include option to specify an alternative ideport that can also be 0 and the system will choose a random port.

Includes #280


src/phpDebug.ts
```diff
@@ -285,7 +285,7 @@ class PhpDebugSession extends vscode.DebugSession {
                         this.sendErrorResponse(response, error)
                         reject(error)
                     })
-                    this._proxyConnect.sendProxyInitCommand()
+                    this._proxyConnect.sendProxyInitCommand(args.port || 9000)
```
felixfbecker: Why not pass this to the `ProxyConnect` like the other config? That would mean you'd have access in the constructor for the message

felixfbecker:
The build is broken (test is failing). That would need to be fixed before merging. 

(I'd like to merge this PR directly instead of the other if we know for a fact that the other one contains a bug)

src/phpDebug.ts
zobo: Personal preference, I guess, did change it however.

zobo:
I'm not sure what's wrong with travis for mac, looks like it can't find php executable...

zobo:
Hey @felixfbecker. I used the better part of the weekend to get this working again. Turns out there was a strange conflict in php versions in the travis macos image. To be honest, I'd much rather just throw macos out of the pipeline. And also bring php and xdebug to a more recent and relevant version.  I'll make another PR for that...

felixfbecker:
Awesome, I'll take another look at the PR and then we can hopefully merge.

src/proxyConnect.ts
```diff
@@ -0,0 +1,132 @@
...
+    duplicateKey: string
+    nonexistentKey: string
+    registerInfo: string
+    registerSuccess: string
+    resolve: string
+    timeout: string
+}
+
+/** Informs proxy of incoming connection and who to pass data back to. */
+export class ProxyConnect extends EventEmitter {
```
felixfbecker: This class has to do very similar things as `XDebugConnection` (manage a socket, parse XML responses, expose methods to send commands), yet doesn't use reuse any of the logic from `DBGPConnection`. This exposes a couple of problems that `DBGPConnection` handles, but `ProxyConnect` does not. Listing out in comments below. The proper solution would likely be that this class should just `extend DBGPConnection` and reuse its logic.

src/proxyConnect.ts
```diff
@@ -0,0 +1,132 @@
...
+        this.emit('info', msg)
+        this._socket.connect(
+            this._port,
+            this._host,
+            () => this._socket.end(cmd)
+        )
+    }
+
+    /** Register/Couples ideKey to IP so the proxy knows who to send what */
+    public sendProxyInitCommand() {
```
felixfbecker: `XDebugConnection` returns a `Promise` from all `send*` methods so that the caller can wait for the response and handle the error. This class just always returns `void`, which for example means when setting up the server, there is no way to wait for the `init` command to have returned (or handle an error).

src/proxyConnect.ts
```diff
@@ -0,0 +1,132 @@
...
+            // Propagate error up
+            this._socket.end()
+            this.emit('error', err instanceof Error ? err : new Error(err))
+        })
+        this._socket.on('lookup', (err: Error | null, address: string, family: string | null, host: string) => {
+            if (err instanceof Error) {
+                this._socket.emit('error', `${err.message}${address || host || ''}`)
+            }
+        })
+        this._socket.on('data', data => this._responseStrategy(data))
```
felixfbecker: This assumes that response messages correspond 1:1 to `data` events, which is not true. It may be true often, especially when testing locally, but `data` events can be arbitrary chunks of data that may contain only part of a response. The `DBGPConnection` class handles this properly.

.travis.yml
```diff
@@ -18,11 +18,8 @@ cache:
     - $HOME/.npm
 
 before_install:
-  - pecl install xdebug-$XDEBUG_VERSION
-  - phpenv config-add travis-php.ini
-  - nvm install $NODE_VERSION
-  - nvm use $NODE_VERSION
-  - npm install -g npm@$NPM_VERSION --depth 0
+  - chmod ugo+x ./scripts/travis-beforeinstall.sh
```
felixfbecker: Why not just `chmod` the checked-in file?

.travis.yml
```diff
@@ -40,26 +37,11 @@ jobs:
...
-        - pecl install xdebug-$XDEBUG_VERSION
-        # make xdebug.so available at pecl's expected location
-        - ln -s /usr/local/Cellar/php@7.1/7.1.26/pecl/ /usr/local/lib/php/pecl
-        - php -i
-        - source ~/.nvm/nvm.sh
-        - nvm install $NODE_VERSION
-        - nvm use $NODE_VERSION
-        - npm install -g npm@$NPM_VERSION --depth 0
+        - chmod ugo+x ./scripts/travis-beforeinstall.sh
+        - ./scripts/travis-beforeinstall.sh 'test'
```
felixfbecker: I'm a bit confused why the subroutine for macOS is called "test" and the one for Linux is called "main"

scripts/travis-beforeinstall.sh
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env bash
+set -x
+
+isFunction() {
```
felixfbecker: This script looks a bit overengineered. How about having two separate scripts so all this logic around argument parsing is not needed?

zobo:
I actually agree with you in a lot of points, it's just that "I already got it like this". I will go over your suggestions and probably rewrite a lot of this code.
How do you feel about macos in travis? I think the scripts can be greatly simplified if we just make use of the provided php images (linux based...). I'd drop macos requirement in travis file.
Thank again for taking the time and being so critical.

felixfbecker:
Not having macOS tests makes me _little_ bit uneasy, but it's been an absolute pain to maintain and configure and Travis takes _forever_ on macOS. One option I was exploring was to use another CI provider with macOS support like Azure DevOps pipelines, but I'm not sure how much that would actually help. I wouldn't be opposed to remove it.

src/proxyConnect.ts
zobo: Although that would be ideal, the reality is, the proxy ide protocol is NOT the same as the dbgp protocol. The specification is loos here and the first major implementation went its own way and now we have a defacto standard. See this issue, where I ask the original author to clarify  this in the specs: https://github.com/derickr/dbgp/issues/19

I will however try to make the class behave more like `XDebugConnection`.

src/proxyConnect.ts
zobo: I agree here, I'll try to make it return a `Promise`. Wather or not to stop on an error is another question.  Currently the implementation shows a warning and continues. I would initially leave it like that, I'll just refactor the plumbing below.

src/proxyConnect.ts
zobo: As said, the IDE registration protocol is not the same as DBGP. The fact is this protocol lacks proper framing of the response (the length NULL xml NULL) so there is no way to know where the response ends. Also the response should always be way below one TCP segment... However, as all DBGP proxy implementations I could find work in a way that they close the TCP session immediately after sending the XML, I have changed the implementation so that it collects all 'data' `Buffer`s and process them on 'close' event.

.travis.yml
zobo: I'm guessing `main` stage and `test` stage, then `release` stage.

.travis.yml
zobo: I'll change it. But I'd rather just throw out macos support and then there's almost no need for this script.. As the Go Proverb goes, "A little copying is better than a little dependency"...

scripts/travis-beforeinstall.sh
zobo: Yop.

.travis.yml
zobo: Did trow out macos and with it this huge script. The few lines that linux needs for initialization are back in the yml.

.travis.yml
zobo: Removed

scripts/travis-beforeinstall.sh
zobo: Removed

elovin:
@zobo there is a bit more active fork of this repository now, maybe you want to continue your work [there](https://github.com/RobberPhex/vscode-php-debug/issues/2) ?

MurzNN:
Is there any progress with merging this PR into this repository, or into some forks like https://github.com/RobberPhex/vscode-php-debug?

zobo:
Sorry about that, I was extremely busy lately. I’ll try to make a for for the other repo.

aldobarr:
@zobo It seems xdebug's dbgpProxy binary is not supported:

https://xdebug.org/docs/dbgpProxy
https://xdebug.org/download#dbgpProxy

zobo:
Reimplemented in #685

PR #388, opened 2020-01-24
Update README.md

wenz:
Derick Rethan*s* :-)


zobo:
Already fixed with #476. Thanks!

PR #405, opened 2020-05-22
feat: alternative source file extensions

mertasan:
I use multiple custom file extensions in my source folder, and these files are automatically build with the .php extension. 

For instance, the configuration is saved in the build folder as **.php** when I process the **.myphp1** **.myphp2** etc. files. 

With these changes, xDebug is identified with srcExtensions items when it cannot find a .php file in its source folder, trying alternative file extensions.


mertasan:
I don't even remember making this PR.

zobo:
Sorry for not processing this earlier, but I could not make sense of it. If you still need it, please provide a more comprehensive use case and we can discuss. Thanks.

PR #421, opened 2020-09-05
Allow for port to be tcp or a unix socket path

moqtadir:
I have been using unix socket paths with vscode-php-debug in a multi-user remote vscode server enivonment. This allows for a clean non-conflicting usage of multiple php debugs with multiple docker containers on same host without having to share tcp ports. 
This is just a typing update so vscode doesn't complain when I use unix path for port.


runephilosof-abtion:
Support for unix sockets seem very useful.
I was looking for a way to use the extension with a unix socket and found this PR.
@felixfbecker Will you comment?

zobo:
Hi!
I will have a look at it shortly, but I don't see a reason not to merge it.

runephilosof-abtion:
@zobo I suggest reopening this PR or creating a follow up issue referencing it, to avoid forgetting it.

zobo:
Hi! I hade this notification on "unread" not to forget, but then personal stuff came by and I did not yet get to it. I will make an issue right away.

PR #472, opened 2020-12-23
build: use ts-node for tests

felixfbecker:
Fixes Codecov


felixfbecker:
Doesn't seem to work for some reason, and #475 fixes codecov.

PR #478, opened 2021-01-04
fix: reset "stopOnEntry" after first use to show correct pause reason

zobo:
If "stopOnEntry" is enabled the pause reason will always be "Paused on entry" event if it's the result of a step operation or breakpoint. Resetting the "stopOnEntry" after first use solves the issue. The flag is not used anywhere else.

### DIFF
--- a/src/phpDebug.ts
+++ b/src/phpDebug.ts
@@ -140,6 +140,9 @@ class PhpDebugSession extends vscode.DebugSession {
     /** A map from unique VS Code variable IDs to XDebug eval result properties, because property children returned from eval commands are always inlined */
     private _evalResultProperties = new Map<number, xdebug.EvalResultProperty>()
 
+    /** A flag to indicate that the adapter has already processed the stopOnEntry step request */
+    private _hasStoppedOnEntry = false
+
     public constructor() {
         super()
         this.setDebuggerColumnsStartAt1(true)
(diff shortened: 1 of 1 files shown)


src/phpDebug.ts
```diff
@@ -365,6 +365,7 @@ class PhpDebugSession extends vscode.DebugSession {
                 exceptionText = response.exception.name + ': ' + response.exception.message // this seems to be ignored currently by VS Code
             } else if (this._args.stopOnEntry) {
                 stoppedEventReason = 'entry'
+                this._args.stopOnEntry = false
```
felixfbecker, line 368: `args` is documented as "The arguments that were given to launchRequest", mutating these at runtime would be confusing if in the future any other part of the code needs to access this and the values don't match the arguments that were given to launchRequest. It would be better to add a separate boolean flag like `hasStoppedOnEntry` that is set to `true` after the first time and also checked.

felixfbecker (changes requested):
Fix seems good except comment I made above.

src/phpDebug.ts
zobo: I have introduced a flag in PhpDebugSession called _hasStoppedOnEntry. It is reset each time before calling the setpInto request.
Now that I think about it more, perhaps it would be better to store this flag into Connection class, as this is a connection specific state?

PR #491, opened 2021-02-02
fix: fixed missing package relateurl

zobo:
Package relateurl was under devDependencies instead of dependencies package.json section.


ajfournier:
Thanks for taking care of this; it was driving me crazy today not knowing why my remote debugging session could no longer be set-up.

Hormah:
So that was unpleasant wasn't it.

PR #525, opened 2021-03-16
fix: Detect IDE path format for convertDebuggerPathToClient

zobo:
Try to detect format of client (IDE) and debugger (Xdebug) path format and explicitly use that instance of path class.


zobo:
This PR also incorporates #365.

PR #536, opened 2021-03-27
chore(release): add marketplace link to release notes

felipecrs:
And also uses the versioned `.vsix` to upload to GitHub Releases.

As in [vscode-shellcheck](https://github.com/timonwong/vscode-shellcheck/blame/3a803d31912b863985405da042339db98fda0c56/release.config.js#L57-L72).

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,13 +82,14 @@
     ],
     "prepare": {
       "path": "semantic-release-vsce",
-      "packageVsix": "php-debug.vsix"
+      "packageVsix": true
     },
     "publish": [
       "semantic-release-vsce",
       {
         "path": "@semantic-release/github",
-        "assets": "php-debug.vsix"
+        "assets": "*.vsix",
+        "addReleases": "bottom"
       }
     ]
   },


zobo:
My first reaction was, "will it work since we don't have a set version in our package.json" (https://github.com/xdebug/vscode-php-debug/blob/main/package.json#L4). Compared to https://github.com/timonwong/vscode-shellcheck/blob/master/package.json#L5. Then I dived into how semantic-release-vsce works and I see it always writes the next version into a local package.json. So this should work. I'll merge it.

My question is tho: How would semantic-version behave if I set a version into out package.json? Will it make a release only when I commit a new version? Do `fix:`, `feat:` and other commit prefixes have any effect on the releasing/publishing flow?

felipecrs:
Yes, in short `semantic-release` completely ignores what you have in your `package.json`, instead it finds the latest version through git tags.

So no matter what you put in the package.json, the rule to trigger releases with semantic-release would still be through commit messages.

If for some reason you need to create a manual release, make sure you create the proper git tag, and semantic-release will be aware of it in the next iteration.

Issue #482, opened 2021-01-11
XDebugError: command is not available


MarceloLuis1997:
Hi,

I'm using this awesome extension for a while and sometimes in the Debug Console it shows the error `XDebugError: command is not available`. The debug still works fine after that, but the Debug Console often shows that error.

PHP version:  7.4.3
XDebug version:  2.9.6-7.4
Adapter version:  v1.14.5

launch.json:  

```
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Listen for XDebug",
            "type": "php",
            "request": "launch",
            "port": 9000
        },
        {
            "name": "Launch currently open script",
            "type": "php",
            "request": "launch",
            "program": "${file}",
            "cwd": "${fileDirname}",
            "port": 9000
        }
    ]
}
```

XDebug php.ini config:

```
zend_extension = C:\xampp\php\ext\php_xdebug-2.9.6-7.4-vc15-x86_64.dll
xdebug.remote_enable = 1
xdebug.remote_autostart = 1
xdebug.remote_host = localhost
xdebug.remote_handler=dbgp
xdebug.remote_port = 9000
```

Error log:

```
connection 342: read ECONNRESET
XDebugError: command is not available
    at new Response (c:\Users\Marcelo\.vscode\extensions\felixfbecker.php-debug-1.14.5\out\xdebugConnection.js:58:19)
    at Connection.<anonymous> (c:\Users\Marcelo\.vscode\extensions\felixfbecker.php-debug-1.14.5\out\xdebugConnection.js:613:20)
    at Generator.next (<anonymous>)
    at fulfilled (c:\Users\Marcelo\.vscode\extensions\felixfbecker.php-debug-1.14.5\out\xdebugConnection.js:5:58)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:94:5) {
  code: 5,
  name: 'XDebugError'
}
connection 363: read ECONNRESET

```

Code snippet to reproduce: It just randomly shows the error in any endpoint of my laravel application.

allanwolski:
Same problem here with my Zend application.

PHP version: 7.1.24
XDebug version: 2.9.8-7.1
Adapter version: v1.14.5

```
XDebugError: command is not available
    at new Response (c:\Users\Allan\.vscode\extensions\felixfbecker.php-debug-1.14.5\out\xdebugConnection.js:58:19)
[...]

PR #550, opened 2021-04-15
feat: Async Breakpoint Manager

zobo:
Handle breakpoints in async manner.
Better support one (VS Code) to many (Xdebug connections).
Handle busy Xdebug connection.


zobo:
Fixes #482

PR #600, opened 2021-06-19
docs: Adding reference to php.validate.executablePath

asopenag:
It wasn't working to me until I learned this step in [this youtube video](https://youtu.be/RXBDt10COsI?t=642)... I think it's useful for dummies like me.


zobo:
Hi! Thanks for this, but I need to look into this as the runtime executable (php.exe) shouldn't come from any VSCode setting. At least not by default...

zobo:
Hi. I revisited this issue, but it does not make much sense. `php.validate.executablePath` is used by php intellisense or the build in php validator, not by this extension.
If it were, there's some other extension manipulating the debugger configuration at runtime.

However I have been recently working on improving the debugging setup experience and have added just this feature. Implemented in #645 users will be able to set `php.executablePath` or `php.validate.executablePath` and if there will be no other php executable defined in `launch.json` one of those will be used.

PR #608, opened 2021-07-19
fix: fixed breakpoint and launch initialization order.

zobo:
Request breakpoints only after LaunchRequest is received, but do not return LaunchResponse until DoneRequest is received. This way all breakpoints will be transferred to DAP and properly mapped.


rceci64:
After this update, the extension does not want to enter into debug mode anymore.
Downgrading to 1.16.1 resolved the problem.

zobo:
Hi! Thanks for the report. Can you give me your php and xdebug version? Are you able to produce any logs by setting  log: true in your launch.json?

zobo:
I was able to reproduce it now. Will look into it asap.

zobo:
Fixed in #608

PR #694, opened 2021-11-23
docs: typo in README.md, installation section php.ini -> xdebug.ini

betovidal:


zobo:
Hi. I'm not sure about this change. Although in many cases you'd have these settings in `xdebug.ini` it can be completely different for other cases. `90-xdebug.ini`, `myconfigs.ini`, `php-7.4.ini` ...
So the docs say `php.ini` that is the root of all configs. And even that can be changed by the distro...

betovidal:
Oh! I see. I had the impression  it had to be in xdebug.ini bacause it had similar configurations and at the end of the setup it worked. Apologies, I haven't used PHP in a while. I'll close this pull request.

PR #763, opened 2022-03-15
fix: send ExitedEvent and TerminatedEvent on process exit

zobo:


JSteitz:
Tested with the same code project, I had the issue.
Looks ok to me, I can terminate the session and dap terminates fine.
Process is also terminated. :+1:

zobo:
Great! Just a warning. In case the DA also starts a php process by adding `program` to launch config (see README) and the debugee is stopped by using the disconnect request, the TerminatedEvent will be sent twice.

One TerminatedEvent comes from when the directly launched debugee exits. This now also sends the ExitedEvent.
The other TerminatedEvent comes from the DisconnectRequest.

The first is needed so that in case the debugee exits normally (for example, the code completed) and there is no DisconnectRequest, the debug session will still terminate.

I tested on VS Code and there were no problems.

I'll batch together this with a few other open things and release it then.

PR #796, opened 2022-04-24
fix: change error message when validating env configuration

zobo:


precha1989 (commented):
#458 #903 #792 #904 #888

PR #882, opened 2023-02-03
feat: Ignore exceptions based on class name

8ctopus:
Add ignore exceptions based on class name.

For some reason, mocha tests don't work on my system. It could be that it's not 100% correct, so please be understanding.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/) and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [1.32.0]
+
+- New launch setting ignoreExceptions.
+
 ## [1.31.0]
 
 - Allow more flexible path mappings in url format.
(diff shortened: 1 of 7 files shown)


zobo:
Hi!

Thanks for the PR. Exception "breakpoints" are a missing feature for some time, but due to the lack of proper UI support they were not yet implemented.

Interesting approach here, I will look at it more closely.

8ctopus:
Thank you, I implemented it this way, because when running Laravel, I get tons of exceptions from within the templating engine, the cookie decryption engine and so forth... This change allows to remove the focus from what can safely be ignored.

zobo:
I completely agree this is needed. I just want to be sure before adding new options to `launch.json` that can be hard to remove later if I decide to reimplement things differently.

zobo:
Try running `npm run lint`

8ctopus:
I ran the linter before implementing the test and didn't check afterwards. It's now fixed. All linter complaints are in parts I didn't touch.

8ctopus:
Any thoughts?

zobo:
Sorry about the delay, I was away. Will take a look at it again in the next days snd probably just merge.

8ctopus:
Thank you for the update @zobo , I can't wait to use it.

zobo:
I did some updates and fixes. Decided to rename the setting to  `ignoreExceptions` since we already have `ignore` and `skipFiles`.

8ctopus:
Awesome, can't wait for the merge. Thank you for looking into it @zobo

8ctopus:
@zobo When can we expect the merge?

zobo:
Wanted to find a better solution as new cases showed up, but can't get to it now, so I merged it now. Hope it helps. Will revisit and find a better solution.

8ctopus:
Awesome thank you.

Issue #890, opened 2023-03-12
ignoreExceptions and backslash errors


zobo:
Test how to properly match namespaced exceptions

PR #891, opened 2023-03-15
fix: namespaced exceptions could not be ignored

8ctopus:
- Namespaced exceptions could not be ignored. I decided to remove minimatch and use a simple comparison instead, because it says `You must use forward-slashes only in glob expressions`.
- Improve ignored exceptions example

### DIFF
--- a/src/phpDebug.ts
+++ b/src/phpDebug.ts
@@ -676,7 +676,7 @@ class PhpDebugSession extends vscode.DebugSession {
                         )) ||
                     // ignore exception class name
                     (this._args.ignoreExceptions &&
-                        this._args.ignoreExceptions.some(glob => minimatch(response.exception.name, glob)))
+                        this._args.ignoreExceptions.some(glob => response.exception.name === glob))
                 ) {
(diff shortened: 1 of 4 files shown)


zobo:
Yes, I noticed and opened an issue for me #890. I'd still like to have wildcard matching so I'm gonna write a few more tests for it.

zobo:
There is also this option: `minimatch(response.exception.name, glob, { windowsPathsNoEscape: true })` but I think it only works if the code runs on windows. So I'll need to find another implementation.

8ctopus:
Actually, I was not properly setup, it works, at least on my OS Windows.

zobo:
I've made a separate PR where I implemented * and ** matching. Please look at https://github.com/xdebug/vscode-php-debug/pull/892/files#diff-df0963a80e4aa05ab41c92cb55f6a3d1e0426a27d746bf5cf3c6ad5bc706e25d for all the cases and let me know if I missed something.

8ctopus:
Sorry, I didn't get a chance to look into it yet.

PR #917, opened 2023-07-22
feat: support clipboard context on variable copy operation

zobo:
When the IDE requests evaluation with clipboard context use var_export to format the response

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/) and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [1.36.0]
+
+- Implement copy to clipboard in var_export format
+
 ## [1.35.0]
 
 - Support for DBGp stream command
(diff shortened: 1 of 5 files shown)


zobo:
Using eval here causes issues, for example, when accessing private properties. Also it does not have support for stack depth. https://bugs.xdebug.org/view.php?id=1342
A correct implementation would be to implement var_export in JS and fetch data via property_get/property_value

eduwass:
I'd love to see this implemented. 👀

almadorx:
Are there any updates on this? It would be definitely great to export all data while debugging

zobo:
Sorry about letting this one fall through the cracks... I'll find some time to implement this asap.

almadorx:
Are there any updates on this? I'm still looking for the possibility of copying all variable values (array, not a single value). Also, it would be great to copy all debug output to a file or clipboard.

zobo:
Not yet, but will get to it. How would you prefer the "copy" operation on a variable would work?
- var_export style?
- json_encode style?
- other?

almadorx:
It doesn't really matter, any variant would be useful, I think.

p.s. I see that you've decided to go with var_export already, great!
 
Thank you!

konigprinz:
Please add this feature

PR #930, opened 2023-11-03
feat: Add limited support for virtual workspaces

zobo:
Override `${workspaceFolder}` variable substitution.


zobo:
Related: https://github.com/microsoft/vscode/issues/197377

PR #943, opened 2023-11-08
fix: Reject any pending commands on connection close

zobo:


MitchTalmadge (approved):
Nice :) The error handling will be awesome in the future.

PR #966, opened 2024-06-30
feat: Add support for runInTerminal request

zobo:
Support for console option with internalConsole, integratedTerminal and externalTerminal options


zobo:
The dilemma here is that it's not trivial to track the spawned process and catch the exit event - and kill the right process.

mature-woman:
Very useful feature!

I use it in NVIM, but the window on the right is always empty ("DAP Console")
изображение

zobo:
Yeah, as said, there are some tricky parts to this. I'll probably implement it in a way that you have to explicitly use it via the `console` setting https://code.visualstudio.com/docs/debugtest/debugging-configuration#_launchjson-attributes

mature-woman:
Thank you

PR #990, opened 2025-02-27
docs: `breakpoint_include_return_value` to known Xdebug config settings

smuuf:
Provide users of this extension with _a knowledge about an option_ to disable the automatic "return value breakpoint" when step debugging.

Xdebug introduced this feature a few years ago, but I find it _extremely_ annoying and difficult to disable. Every time I set up a new PHP development environment in vscode, I have to rediscover how to turn it off.

The configuration option `breakpoint_include_return_value` is hard to find, as it isn’t even mentioned in the official [Xdebug documentation](https://xdebug.org/docs/). The only way to learn about it is by venturing into [Xdebug’s C source code](https://github.com/search?q=repo%3Axdebug%2Fxdebug+breakpoint_include_return_value&type=code).

To improve usability, let’s add this setting to the known options available under the `xdebugSettings` key.


zobo:
Hi! Thanks for the PR. I believe I never added it to docs, as it's not an official setting yet. However in contrast, I did enable the feature by default.... so there's a strong argument to add this.

Please give me a few days to test this out as I'm currently traveling.

PR #1017, opened 2025-04-25
chore: Disable OVSX_PAT

zobo:


Unradelic:
I don't know how or why, but this commit broke the extension for Linux upon debug session start. The error `Exception has occurred: Error: Unknown sourceReference 0` keeps appearing in my Linux machines with PoP_OS and CachyOS. This issue doesn't happen in WSL2 (because, I presume it uses Ubuntu 22)

image

I have tested debugging the extension and checking out to the previous commit (#c954087
) and this problem doesn't appear anymore.

Out of curiosity, what is and what does OVSX_PAT? Why was this ENV removed?

